PR #1, opened 2017-09-26
RPS-16 Log in - code review request

ziemsky:
Hi, Ivor.

Here is the 'RPS-16 Log in' story ready for your review.

As discussed, I've kept the 'vanilla project' commit separate but all the later commits have been squashed together. I've also corrected the test failure messages in the way we agreed.

Cheers,
Rafał


README.txt
```diff
@@ -0,0 +1,93 @@
...
+  mvn clean verify -Djrebel
+
+Start with:
+
+  mvn -P cargo.run -Djrebel
+
+Best Practice for development
+=============================
+
+Use the option -Drepo.path=/some/path/to/repository during start up. This will avoid
```
ivorc: What will be stored in the repo? Can you just rm -rf it if you want to? What happens? When would/wouldn't you want to do such a thing?

README.txt
```diff
@@ -0,0 +1,93 @@
...
+  mvn -P cargo.run -Drepo.path=/home/usr/tmp/repo
+
+or with jrebel:
+
+  mvn -P cargo.run -Drepo.path=/home/usr/tmp/repo -Djrebel
+
+Hot deploy
+==========
+
+To hot deploy, redeploy or undeploy the CMS or site:
```
ivorc: So do you run mvn -Pcargo.run in one shell and mvn cargo:redeploy in another? Maybe just explain a little more.

README.txt
```diff
@@ -0,0 +1,93 @@
...
+
+  cd cms (or site)
+  mvn cargo:redeploy (or cargo:undeploy, or cargo:deploy)
+
+Automatic Export
+================
+
+Automatic export of repository changes to the filesystem is turned on by default. To control this behavior, log into
```
ivorc: What does this mean in practice? What's the workflow? (I guess it's to git add the new/modified files. In all, or only some, cases?)

README.md
```diff
@@ -4,3 +4,39 @@
 
 To quickly setup this project run `make init`. To get more help, simply run
 `make help`.
+
```
ivorc: Nice 👍

acceptance-tests/pom.xml
```diff
@@ -0,0 +1,362 @@
...
+        <dependency>
+            <groupId>javax.mail</groupId>
+            <artifactId>mail</artifactId>
+        </dependency>
+
+    </dependencies>
+
+    <build>
+        <plugins>
+            <!--
```
ivorc: Nice comments 👍

acceptance-tests/src/test/java/uk/nhs/digital/ps/test/acceptance/pages/AbstractPage.java
```diff
@@ -0,0 +1,23 @@
...
+import org.openqa.selenium.WebDriver;
+import uk.nhs.digital.ps.test.acceptance.webdriver.WebDriverProvider;
+
+/**
+ * Base class for classes implementing 'Page Object' pattern; provides common properties (such as base URL)
+ * and dependencies (such as web driver) to extending classes.
+ */
+abstract class AbstractPage {
+
+    static final String URL = "http://localhost:8080/cms";
```
ivorc: Might be nice to have this as a config parameter to allow it to be easily changed on systems where 8080 is already in use. Is that relatively easy?

acceptance-tests/src/test/java/uk/nhs/digital/ps/test/acceptance/pages/LoginPage.java
```diff
@@ -0,0 +1,63 @@
...
+        getWebDriver().get(URL);
+    }
+
+    public void loginWith(final String username, final String password) {
+        getUsernameField().sendKeys(username);
+        getPasswordField().sendKeys(password);
+        getLoginButton().click();
+    }
+
+    private WebElement getLoginButton() {
```
ivorc: It's a small thing, but calling this "find", rather than "get", would more clearly indicate the behaviour to me. i.e. it's not just returning a value that's already stored, but instead going out and hunting it down. Same comment for other equivalent "get" methods.

acceptance-tests/src/test/java/uk/nhs/digital/ps/test/acceptance/pages/LoginPage.java
```diff
@@ -0,0 +1,63 @@
...
+
+    private WebElement getPasswordField() {
+        return getWebDriver().findElement(By.name("password"));
+    }
+
+    private WebElement getUsernameField() {
+        return getWebDriver().findElement(By.name("username"));
+    }
+
+    public boolean hasLoginErrorMessage(String expectedErrorMessage){
```
ivorc: Is this used? Looks like it's not.

acceptance-tests/src/test/java/uk/nhs/digital/ps/test/acceptance/steps/LoginSteps.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    @When("^I submit my invalid credentials$")
+    public void iSubmitMyInvalidCredentials() throws Throwable {
+        loginPage.loginWith("admin","wrongpassword");
+    }
+
+    @Then("^an error is displayed$")
+    public void anErrorIsDisplayed() throws Throwable {
+        // @formatter:off
```
ivorc: Do we really need to disable formatting?

README.md
ziemsky: Credit for this bit goes to Wojtek :-)

wojtek-oledzki:
@ziemsky squash to one commit please.

wojtek-oledzki:
@ziemsky can you delete all .class files please

wojtek-oledzki:
@ziemsky do we need the "essentials/" folder in the repo?

acceptance-tests/src/test/java/uk/nhs/digital/ps/test/acceptance/pages/AbstractPage.java
ziemsky: Indeed, in the future it may be more useful to externalise this URL, although it may also turn out that we never need to do it, so I decided to not to spend time on implementing something we may never need. The default configuration included in the project as generated by Hippo's Maven prototype does the job for now.

Should we need to make this value configurable and, say, have it set to different values depending on the environment, 
it will be reasonably easy to do so - given that the tests' configuration is being managed by Spring we can easily leverage the wealth of the various methods for externalising configuration that Spring supports - properties files, command line arguments, system/environment properties, values passed in from Maven.

acceptance-tests/src/test/java/uk/nhs/digital/ps/test/acceptance/pages/LoginPage.java
ziemsky: Good point. I'll rename the methods.

acceptance-tests/src/test/java/uk/nhs/digital/ps/test/acceptance/pages/LoginPage.java
ziemsky: Indeed, I left it behind by mistake having done refactoring that rendered this method unused. Gone now.

acceptance-tests/src/test/java/uk/nhs/digital/ps/test/acceptance/steps/LoginSteps.java
ziemsky: I thought this long statement reads better when broken down like this and therefore wanted to ensure it remains that way in case someone runs auto-formatter against the whole file.

Having said that, I'm not fussed about it here - the line is simple enough that even when it breaks down after the second comma (which happens when default auto-formatting is applied) it still remains sufficiently readable. I'm removing the auto-formatter statements.

ziemsky:
Hi, @wojtek-oledzki:

**Re: squashing:**
Commit `ebe6695` submitted for code review is a number of commits squashed into one, already. It's preceded by three commits, where the last of them - _`9f13513` [RPS-16] Create vanilla Hippo project_ - has been left intact as agreed with Ivor so that we have one, distinct commit that captures the state of the project as it's been right after it's been generated from Hippo Maven archetype (that is without our changes). This is so that we can easily go back to the starting point and compare against our later changes should things start going wrong. A consequence of leaving this commit to stand alone is that I can't squash any commits that precede it.

**Re: removing Essentials**
In short: I'm not absolutely sure whether we do or do not need this module, yet, so I decided to play safe and leave it in until we learn more and can make a more informed decision.

The Essentials application allows us to add features to the project that could still prove needed, especially the ones classed as 'Enterprise'. We can delete this module at any time, and, until then, it shouldn't really get in the way - at least it didn't so far.

**Re: removal of the *.class files**
I absolutely agree that it stinks to have these files in the codebase.

These files were generated automatically by Hippo Maven archetype and I'm guessing there was a reason why they were placed in the part of the source code that gets checked into the VCS.

All the `*.class` files that I was able to find belong to the Essentials module and all the uncertainty about whether we need the module or not, apply to these files as well. Admittedly, when I deleted them, they did seem to get re-generated under `target/essentials/WEB-INF/classes` directory but they differed from the originals in a very subtle way - the regenerated files were missing a few files with suffix `$1` that could be find in the deleted original content. Even though the Essentials application seemed to be working okay in my very brief test, I remain wary of the (small, I think) risk that some problems may arise later due to our intervention and the debugging can be difficult because of the time passing between the deletion and the problems being encountered.

Again, ~being insanely paranoid~ just to be on the safe side, while not being happy to keep any sort of binary files in the repo, I opt to leave these files as they are. Given that they collectively take up less than 300k I think we won't feel any impact.

README.md
```diff
@@ -4,3 +4,39 @@
...
+## Key build operations
+
+This project uses Maven as its build system. Unless stated otherwise, instructions below assume that Maven commands
+will be executed from the directory of the parent module.
+
+To run:
+
+**Unit tests**:
+> mvn test
```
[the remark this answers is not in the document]
ziemsky: Indeed, the 'code blocks' do read better than the 'quote blocks' - replacing now.

wojtek-oledzki:
@ivorc @ziemsky for simplicity - https://github.com/NHS-digital-website/ps-hippo/pull/2 vanila hippo.

Let's merge #2 first and then come back to this PR.

PR #3, opened 2017-09-28
[RPS-56] Run acc. tests against standalone Hippo instance

ziemsky:
Through the use of Maven profiles 'cargo.run', 'acceptance-test' and
'acceptance-test-hippo', it is now possible to run automated acceptance
tests against an instance of Hippo CMS that was either started manually,
running constantly in the background, or one that is started and
stopped automatically prior and after the tests, respectively.

The details of how to use the profiles have been described in the
readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,49 +5,103 @@
 To quickly setup this project run `make init`. To get more help, simply run
 `make help`.
 
+## Key Maven Build Operations
 
+This project uses [Maven] as its build system. Unless stated otherwise, instructions below assume that Maven commands
+will be executed from command line in the directory of the parent module.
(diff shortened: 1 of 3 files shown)


ziemsky:
TL;DR: previously, Maven Cargo plugin (runs Tomcat and deploys apps) and Failsafe plugin (runs acceptance tests) were configured as part of the same Maven profile (acceptance-test) which caused them to always run together. Moving configuration of the Cargo plugin into a separate profile 'acceptance-test-hippo' enables them to be run independently, so that we don't have to cycle the server every time we want to run the tests against it.

PR #10, opened 2017-10-06
RPS-55 Make information type mandatory

ivorc:


acceptance-tests/src/test/java/uk/nhs/digital/ps/test/acceptance/pages/ContentPage.java
```diff
@@ -45,6 +45,10 @@ public void populatePublication(Publication publication) {
 
         WebElement summaryField = editorBody.findElement(By.className("publication-summary")).findElement(By.tagName("textarea"));
         summaryField.sendKeys(publication.getPublicationSummary());
+
+        WebElement informationTypeWebElement = editorBody.findElement(By.xpath("//span[text()='Information Type']/../following-sibling::div//select[@class='dropdown-plugin']"));
+        Select informationType = new Select(informationTypeWebElement);
+        informationType.selectByIndex(1);
```
ngta1: Do we need to select by text i.e "Official statistics" and then update the front-end check to ensure "Official statistics" is actually displayed (similar to what we did for title and summary)?  Will bloat the tests if so, therefore could be overkill for something as simple as a selecting a dropdown.

ngta1: Thinking more about this...selecting by Index may actually be more future-proof since if we did select by text, and "Official statistics" was ever removed in the future, we would have to update our test code as well.

ivorc:
Agreed, added assertion on Information Type being rendered as expected.

ngta1:
All good to go.

PR #15, opened 2017-10-12
RPS-20 file types rejection test

ziemsky:
Hi, Ivor.

Here is my implementation of the 'negative' file upload test - the one that ensures that files from outside of whitelist is rejected.

Once reviewed and approved **PLEASE DO NOT MERGE THIS BRANCH YET** - this branch is based on `RPS-20_file_types_acceptance_test` which still needs post-code review amendments applied first (next thing on my todo). Once that is done, I'll rebase this one and only then we'll merge it to master.


acceptance-tests/src/test/java/uk/nhs/digital/ps/test/acceptance/pages/AbstractCmsPage.java
```diff
@@ -0,0 +1,25 @@
...
+
+    static protected final String URL = "http://localhost:8080/cms";
+
+    AbstractCmsPage(final WebDriverProvider webDriverProvider) {
+        super(webDriverProvider);
+    }
+
+    public void logout() {
+
+        // @formatter:off
```
ivorc: Just wondering why the formatter:off is needed? Seems to format OK in IntelliJ for me.

acceptance-tests/src/test/java/uk/nhs/digital/ps/test/acceptance/steps/ContentSteps.java
```diff
@@ -102,52 +114,100 @@ public void whenIPublishThePublication() throws Throwable {
...
+
+    @Then("^the upload is rejected with an error message$")
+    public void theUploadIsRejectedWithAnErrorMessage() throws Throwable {
+
+        testDataRepo.getCurrentAttachments().forEach(attachment -> {
+            contentPage.getAttachmentsSection().uploadAttachment(attachment);
+
+            final String expectedErrorMessage = MessageFormat.format(
+                "The uploaded file ''{0}'' has extension {1} which is not allowed." +
+                    " Allowed extensions are: doc, docx, xls, xlsx, pdf, csv, zip, txt, rar, ppt, pptx.",
```
ivorc: Suggest we take this list of extensions out, since this may get out of date, and I don't think it's necessary anyway if we can match on what the string *starts with*

acceptance-tests/src/test/java/uk/nhs/digital/ps/test/acceptance/steps/WebDriverSteps.java
```diff
@@ -19,7 +19,8 @@ public void initialise() {
         webDriverProvider.initialise();
     }
 
-    @After
+    // Low execution order ensures that the web driver is disposed after any other @After steps that may need it.
+    @After(order = 0)
```
ivorc: Oooh, clever 👍

acceptance-tests/src/test/java/uk/nhs/digital/ps/test/acceptance/pages/AbstractCmsPage.java
ziemsky: Force of habit - we used auto-formatting a lot on my previous project and we'd often exclude this type of expressions from its scope like this, to preserve readability.

Since we don't use auto-formatting here, there is no need for these `// @formatter:off`/`// @formatter:on` statements - I have now removed them.

acceptance-tests/src/test/java/uk/nhs/digital/ps/test/acceptance/steps/ContentSteps.java
ziemsky: Good point. The static list is now removed and the matcher switched to use 'startsWith`.

ziemsky:
**DO NOT MERGE THIS BRANCH** - it has been incorporated into `RPS-20_file_types_rejection_test` and it's the latter that will be merged into `master`.

PR #26, opened 2017-10-28
[RPS-65] Add publication summary validation

wojtek-oledzki:


acceptance-tests/src/test/java/uk/nhs/digital/ps/test/acceptance/steps/ContentSteps.java
```diff
@@ -266,30 +268,34 @@ public void iCanSeeAListOfPublicationsFromTheTimeSeries() throws Throwable {
...
         }
         String longString = sb.toString();
-        assertThat("String should be > 250 characters long", longString.length(), greaterThan(250));
 
-        contentPage.populatePublicationTitle(longString);
+        if (fieldName.equals("title")) {
+            contentPage.populatePublicationTitle(longString);
+        } else if (fieldName.equals("summary")) {
+            contentPage.populatePublicationSummary(longString);
+        }
```
ziemsky: It may be just me, but I'd split this into two, separate method and eliminate the if/else logic, while moving the string generation into a common helper/utility method.

Admittedly, right now it's a rather simple case, and in tests it may have even less impact than in production code, but an initially innocent looking if/else logic is prone to quickly become difficult to maintain as we may be tempted to add cases to it; in general, the less conditional logic the easier to read and maintain (and test, in case of production code) the code is so it's best to avoid it if we can.

acceptance-tests/src/test/java/uk/nhs/digital/ps/test/acceptance/steps/ContentSteps.java
```diff
@@ -266,30 +268,34 @@ public void iCanSeeAListOfPublicationsFromTheTimeSeries() throws Throwable {
             is(expectedPublicationTitles));
     }
 
-    // Scenario: Title validation ========================================================================
-    @When("^I populate the title with text which is too long$")
-    public void whenIPopulateTheTitleWithTextWhichIsTooLong() throws Throwable {
+    // Scenario: Title and Summary validation ========================================================================
+    @When("^I populate the (title|summary) with text which is ([0-9]+) characters long$")
```
ziemsky: How about rephrasing this to say something along the lines of _'I populate the (title|summary) with text which exceeds maximum allowed limit of ([0-9]+) characters long'_?

I think this would restore the expressiveness of the original version of this step, while retaining ability to specify the actual number.

acceptance-tests/src/test/java/uk/nhs/digital/ps/test/acceptance/steps/ContentSteps.java
```diff
@@ -266,30 +268,34 @@ public void iCanSeeAListOfPublicationsFromTheTimeSeries() throws Throwable {
...
 
     @When("^I save the publication$")
     public void iSaveThePublication() throws Throwable {
         contentPage.savePublication();
     }
 
-    @Then("^title validation error message is shown$")
-    public void titleValidationErrorMessageIsShown() throws Throwable {
+    @Then("^validation error message is shown and contains \"([^\"]+)\"$")
+    public void validationErrorMessageIsShownAndContains(String errorMessageFragment) throws Throwable {
```
ziemsky: Nice, clean re-use thanks to introduction of the parameter :+1:

acceptance-tests/src/test/java/uk/nhs/digital/ps/test/acceptance/steps/ContentSteps.java
ziemsky: We all appreciate the situation here and agreed that just adding a clarifying comment will suffice for now.

PR #29, opened 2017-10-31
[RPS-132] Prevent saving publication with blank Granularity field

ziemsky:
Removed Granularity from:
- publication.yaml prototypes so that an empty placeholder doesn't
  get automatically added to newly created documents,
- from the test 'lorem ipsum' documents.

This is a change missed from the main commit 4517457.


wojtek-oledzki:
Looks good to me :tm:

LGTM

PR #39, opened 2017-11-15
[RPS-137] Add test users

wojtek-oledzki:
This commit also contains tiny cleanup of the SearchComponent (thanks
@wbarthet)


ci-cd/bin/get-hippo-groups-yaml
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+
+DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
```
ivorc: Could do with some commenting, I think. It may be just because I'm not a bash expert, but I find it hard to follow this.

ci-cd/bin/get-hippo-groups-yaml
```diff
@@ -0,0 +1,36 @@
...
+
+function main {
+    local authors="      - author"
+    local editors="      - editor"
+
+    while read -r line; do
+        id=$(echo "${line}" | sed -e 's/ /./' | tr '[:upper:]' '[:lower:]')
+        authors="${authors}
+      - ${id}.a"
+    done < "$1"
```
ivorc: Worth storing $! in a local variable to be more self-documenting?

PR #44, opened 2017-11-23
[RPS-167] wrong resources file type

jacsamell:
Removed the resource type indicators for links and files on the publication and the dataset template. Leaving an indicator for datasets on the publication template to distinguish these from links.

Also removed a trailing 'l' that was getting incorrectly added to the end of the publication title.


wojtek-oledzki:
Pure Awesomeness!

Pure Awesomeness

PR #49, opened 2017-11-29
[RPS-171] Move getParentPublication to Dataset bean

jacsamell:
Moved from the DatasetComponent to make more object oriented

### DIFF
--- a/site/src/main/java/uk/nhs/digital/ps/components/DatasetComponent.java
+++ b/site/src/main/java/uk/nhs/digital/ps/components/DatasetComponent.java
@@ -22,7 +22,7 @@ public class DatasetComponent extends BaseHstComponent {
         super.doBeforeRender(request, response);
         final HstRequestContext ctx = request.getRequestContext();
         Dataset dataset = (Dataset) ctx.getContentBean();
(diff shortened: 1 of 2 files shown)


site/src/main/java/uk/nhs/digital/ps/components/DatasetComponent.java
```diff
@@ -22,7 +22,7 @@ public void doBeforeRender(final HstRequest request, final HstResponse response)
         super.doBeforeRender(request, response);
         final HstRequestContext ctx = request.getRequestContext();
         Dataset dataset = (Dataset) ctx.getContentBean();
-        Publication publication = getParentPublication(ctx, dataset);
+        Publication publication = dataset.getParentPublication(ctx);
```
wojtek-oledzki: sugestion: pass "rootFolder" (`ctx.getSiteContentBaseBean()`) instead of ctx.

jacsamell: I'm changing this again for RPS-170 so will merge this change into that and submit pull requests together

PR #55, opened 2017-11-30
[RPS-181] Remove attachment body from search scope

wojtek-oledzki:
This is a test approach, that uses derived content mechanism to
overwrite "hippo:text" property with empty string which makes given
resource non-searchable.

NOTE: Attachment file name is still searchable.

### DIFF
--- a/repository-data/application/src/main/resources/hcm-config/namespaces/publicationsystem/publication.yaml
+++ b/repository-data/application/src/main/resources/hcm-config/namespaces/publicationsystem/publication.yaml
@@ -310,6 +310,7 @@ definitions:
             plugin.class: org.hippoecm.frontend.editor.plugins.field.NodeFieldPlugin
             wicket.css: [attachments]
             wicket.id: ${cluster.id}.left.item
+            searchable: false
             /cluster.options:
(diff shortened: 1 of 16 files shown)


repository-data/application/src/main/resources/hcm-config/namespaces/publicationsystem/publication.yaml
```diff
@@ -115,8 +115,8 @@ definitions:
             hipposysedit:ordered: false
             hipposysedit:path: publicationsystem:attachments
             hipposysedit:primary: false
-            hipposysedit:type: hippo:resource
-            hipposysedit:validators: [publicationsystem-blank-attachment]
+            hipposysedit:type: publicationsystem:resource
+            # hipposysedit:validators: [publicationsystem-blank-attachment]
```
wojtek-oledzki: ups - I need to update validator class

wojtek-oledzki:
Another way is to copy ResourceUploadPlugin class and simply not export "hippo:text" for PDFs.

wojtek-oledzki:
Closing for now - trying something else

PR #56, opened 2017-11-30
[RPS-184] Prepare project for .less to .css compilation

jacsamell:
Added maven plugin to compile less files to the css folder where they are picked up by hippo.
Removed .css files from git and replaced with .less equivalent.


jacsamell:
Changes made.

PR #63, opened 2017-12-04
[RPS-173] Added a script to format yaml files

jacsamell:
Note: There are 2 commits with this, the first is the functional change to add the script and the second is running the script on our YAML files to get them in the right format.

The groovy script orders the entries in the yaml files alphabetically and removes any UUIDs.
This should reduce changes when using auto-export so diffs will show only what has actually changed.
The script will run on maven verify and I've also added a command to run it manually with make from the command line.

Also performed an initial run of YAML formatter to write files in the new standard format.
In future all diffs for these files will be only the functional changes.
There are no functional changes to the YAML, the main formatting differences are:
 - Entries are sorted alphabetically
 - All files start with ---
 - All arrays are now formatted as lists
 - Long key values are wrapped and prefixed with '?' (See complex keys in YAML docs)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,8 +12,13 @@ help:
 ## Initialise local project
 init: .git/.local-hooks-installed
 
-## Clean, build and start local hippo (no autoexport)
+## Clean, build and start local hippo
 serve:
+	mvn clean verify
+	mvn -P cargo.run
+
+## Serve without allowing auto-export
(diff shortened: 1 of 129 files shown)


woledzki:
I will checkout the branch and test locally.

wojtek-oledzki:
I will checkout the branch and test locally.

PR #64, opened 2017-12-04
[RPS-150] Basic formatting of summary field

wojtek-oledzki:
Summary field in a) Series, b) Publicatin and c) Dataset can now contain
multiple paragraphs separated with 2 or more breaklines.


site/src/main/java/uk/nhs/digital/ps/beans/Dataset.java
```diff
@@ -66,6 +68,10 @@ public Calendar getCoverageEnd() {
         return getProperty("publicationsystem:CoverageEnd");
     }
 
+    public List<Element> getSummaryParagraphs() {
+        return getParsedText("publicationsystem:Summary");
+    }
```
ziemsky: Publication now has two getters that return the same content in two different formats - `getSummary` and
`getSummaryParagraphs`, each returning the same content in two different formats.

In recognition that it's difficult to completely divorce content of summary from all formatting concerns (we do want to
let the users apply paragraphs and, later, possibly list of bullet points), how does it soud if we:
a) introduced a class `StructuredText` to encapsulate collection of paragraphs,
b) made `Publication.getSummary` return instance of that class,
c) moved `Publication.getSummaryParagraphs` to `StructuredText`, renaming it to something like `getElements`,
  `getComponents` or `getParts`,
d) moved `Publiction.parseText` to `StructuredText`.

This way all low level manipulations related to text of this particular format would be encapsulated in a class whose
only concerns would be just that - to hold and correctly process a structured text rather than be concerned what's in
it. Conversely, all that `Publiction` would know is that it holds and returns a Summary - without being burdened with
details of how to process it.

`StructuredText` could hold the raw text and parse it whenever its getters are called, or parse it once, when it's being
created, and store collection of Elements. It could also expose `getFirstNCharacters(int charactersCount)` for where
the summary needs to be displayed truncated in `publications-overview.ftl` and in `searchresults.ftl` (if/when we later
add support for more types of elements, say lists of bullet points, this will allow to intelligently encapsulate logic
for correct extraction of first characters - for example, what if the first N characters spill beyond first paragraph
and into the list of bullet points?).

site/src/main/java/uk/nhs/digital/ps/beans/Publication.java
```diff
@@ -236,6 +237,11 @@ public boolean isPubliclyAccessible() {
         return getChildBeansIfPermitted(PropertyKeys.ATTACHMENTS, HippoResourceBean.class);
     }
 
+    public List<Element> getSummaryParagraphs() {
+        assertPropertyPermitted(PropertyKeys.SUMMARY);
+        return getParsedText(PropertyKeys.SUMMARY);
+    }
```
ziemsky: Publication now has two getters that return the same content in two different formats - `getSummary` and
`getSummaryParagraphs`, each returning the same content in two different formats. `BaseDocument` has gained `getParsedText` which isn't even needed by all of the extending classes. Both these additions mean that `Publication` (and other classes extending `BaseDocument`), apart from being concerned with just the content are now also burdened with formatting concerns which normally are of interest to the presentation/view layer rather than domain layer.

In recognition that it's difficult to completely divorce content of summary from _all_ formatting concerns (after all, we do want to let the users apply paragraphs and, later, possibly list of bullet points, and we do have to persist the information of this structure somehow), how does it sound if we:
1) introduced a class `StructuredText` to encapsulate collection of paragraphs,
1) made `Publication.getSummary` return instance of that class (the same for other classes extending `BaseDocument`),
1) moved `Publication.getSummaryParagraphs` to `StructuredText`, renaming it to something like `getElements`,
  `getComponents` or `getParts`,
1) moved `BaseDocument.parseText` to `StructuredText`.

This way, all low level manipulations related to text of this particular text format would be encapsulated in a class whose only concerns would be just that - to hold and correctly process a structured text rather than be concerned what's in it. Conversely, all that `Publication` would know is that it holds and returns a Summary - without being burdened with details of how to process it.

`StructuredText` could hold the raw text and parse it whenever its getters are called, or parse it once, when it's being created, and store collection of Elements. It could also expose `getFirstNCharacters(int charactersCount)` for where the summary needs to be displayed truncated in `publications-overview.ftl` and in `searchresults.ftl` (if/when we later add support for more types of elements, say lists of bullet points, this will allow to intelligently encapsulate logic for correct extraction of first characters - for example, it'll help dealing with the case when the first N characters spill beyond the first paragraph and into the list of bullet points).

wojtek-oledzki:
@ziemsky all applied.

PR #68, opened 2017-12-05
[RPS-190] Update Information Type to non-mandatory

ngta1:


repository-data/application/src/main/resources/hcm-config/namespaces/publicationsystem/publication.yaml
```diff
@@ -278,9 +278,6 @@ definitions:
             hipposysedit:path: publicationsystem:InformationType
             hipposysedit:primary: false
             hipposysedit:type: StaticDropdown
-            hipposysedit:validators:
-            - non-empty
```
ziemsky: As discussed, how about following example set in Granularity and making it so that if no value is selected, then no blank field is displayed, requiring the user to click 'Add' before adding even a single value.

If possible, it would then be good to leave the `non-empty` validator in place to make sure that users don't submit blank fields.

Note that the `/prototype` section also needs updating to ensure that newly created documents do not have empty fields by default.

ngta1: Good spot.  Will correct.

wojtek-oledzki:
Looks good to me :tm:

LGTM

PR #69, opened 2017-12-06
[RPS-160] Clinical indicator summary migration

jacsamell:
Added summary migration for clinical indicators.
Also removed the Purpose field from the CMS as we want the purpose to
 just be the last paragraph in the summary.
Also small bit of refactoring to remove duplicate TaxonomyName method.

### DIFF
--- a/migrator/src/main/java/uk/nhs/digital/ps/migrator/task/ImportableItemsFactory.java
+++ b/migrator/src/main/java/uk/nhs/digital/ps/migrator/task/ImportableItemsFactory.java
@@ -38,7 +38,8 @@ public class ImportableItemsFactory {
         return new DataSet(
             parentFolder,
             exportedPublishingPackage.getTitle(),
-            exportedPublishingPackage.getTitle()
+            exportedPublishingPackage.getTitle(),
+            exportedPublishingPackage.getSummary()
         );
     }
 
--- a/migrator/src/main/resources/templates/publication.json.ftl
+++ b/migrator/src/main/resources/templates/publication.json.ftl
@@ -27,7 +27,7 @@
     "name" : "publicationsystem:Summary",
     "type" : "STRING",
     "multiple" : false,
-    "values" : [ "Some summary" ]
+    "values" : [ "Publication" ]
   }, {
     "name" : "publicationsystem:CoverageEnd",
     "type" : "DATE",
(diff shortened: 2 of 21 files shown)


wojtek-oledzki:
I will re-run migration locally and check output later today.

site/src/main/java/uk/nhs/digital/common/components/FacetComponent.java
```diff
@@ -7,35 +7,20 @@
...
+
 import javax.jcr.RepositoryException;
 
 public class FacetComponent extends SearchComponent {
 
     @Override
     public void doBeforeRender(HstRequest request, HstResponse response) {
         TaxonomyManager taxonomyManager = HstServices.getComponentManager().getComponent(TaxonomyManager.class.getName());
-        Taxonomy taxonomy = taxonomyManager.getTaxonomies().getTaxonomy(getTaxonomyName(request.getRequestContext()));
+        Taxonomy taxonomy = taxonomyManager.getTaxonomies().getTaxonomy(HippoBeanHelper.getTaxonomyName());
```
ngta1: Nice 👍

migrator/src/main/java/uk/nhs/digital/ps/migrator/task/ImportableItemsFactory.java
```diff
@@ -38,7 +38,8 @@ public static DataSet toDataSet(final PublishingPackage exportedPublishingPackag
         return new DataSet(
             parentFolder,
             exportedPublishingPackage.getTitle(),
-            exportedPublishingPackage.getTitle()
+            exportedPublishingPackage.getTitle(),
```
ngta1: Duplicate getTitle?

migrator/src/main/resources/templates/publication.json.ftl
```diff
@@ -27,7 +27,7 @@
     "name" : "publicationsystem:Summary",
     "type" : "STRING",
     "multiple" : false,
-    "values" : [ "Some summary" ]
+    "values" : [ "Publication" ]
```
ngta1: Does this need to be more specific to denote it is a summary value?

migrator/src/main/java/uk/nhs/digital/ps/migrator/task/ImportableItemsFactory.java
jacsamell: At the moment the dataset name and title are both set to the same thing as we only have this value from the exported CI data.

migrator/src/main/resources/templates/publication.json.ftl
jacsamell: This is just place holder text for the top level publication (e.g. Dec-2017) as there are no summaries to migrate from the data for publications. This will be changed in the CMS under the 'Summary' header by SMEs so I think it's ok for now if you agree? I was trying to improve it from 'some summary' 🙂

PR #71, opened 2017-12-06
[RPS-184] Styleguide new designs

dsayce:
Applied styling according to the design.

Also, removed redundant test for summary length validation.

All the Java changes are from @ziemsky


repository-data/webfiles/src/main/resources/site/less/demo.less
```diff
@@ -1,303 +1,458 @@
...
-    background-repeat: no-repeat;
-    text-indent: -9999em;
+
+
+
+
+/* HEADER */
+/* ########################################################## */
+
+.header {
```
ziemsky: As discussed, can we add some qualifier to this, renaming it to something like 'top-header'?

repository-data/webfiles/src/main/resources/site/less/demo.less
```diff
@@ -1,303 +1,458 @@
...
+@base-radius: 5px;
 
 
-/*
-<div class="container">
-</div>
- */
 .container {
-    width: 1200px;
+    width: @max-width;
```
ziemsky: As discussed, can we add some qualifier to this, renaming it to something like 'document-content'?

repository-data/webfiles/src/main/resources/site/less/demo.less
```diff
@@ -1,303 +1,458 @@
...
-    margin-bottom: 20px;
-    overflow: hidden;
-    text-align: center;
+
+
+/* SEARCH */
+/* ########################################################## */
+
+.search {
+  background-color: @color-secondary;
```
ziemsky: As discussed, can we add some qualifier to this and derived classes, renaming it to something like 'common-search'?

styleguide/index.html
```diff
@@ -0,0 +1,96 @@
...
+    <ul>
+      <li><a href="http://localhost:8080/publications/acceptance-tests/series-with-publication-with-datasets/series-publication-with-datasets/series-publication-with-datasets-dataset">Example page</a>
+      <li><a href="../styleguide/dataset.html">Template</a></li>
+    </ul>
+  </div>
+
+  <div class="panel panel--grey push--bottom">
+    <h3>Publication:</h3>
+    <ul>
+      <li><a href="http://localhost:8080/publications/acceptance-tests/publication-with-datasets">Example page</a>
```
ziemsky: I think this URL needs changing to relative paths.

repository-data/webfiles/src/main/resources/site/less/demo.less
wojtek-oledzki: `container` is pretty well defined in Frontend Land. I would be surprised not to find "container" class tbh.

styleguide/dataset.html
```diff
@@ -0,0 +1,166 @@
+<!DOCTYPE html>
+<html lang="en" class="">
+<head>
+
+  <title>NHS - Replacement Publication System - Dataset</title>
+
+  <meta charset="UTF-8" />
+  <meta name="title" content="NHS - Replacement Publication System" />
+  <meta name="viewport" content="width=device-width, minimum-scale=1.0, maximum-scale=1.0" />
+  <link rel="stylesheet" href="../repository-data/webfiles/src/main/resources/site/css/demo.css" />
```
ziemsky: This path appears to be pointing to a file that is dynamically generated from `demo.less` and, unless the build is run, the `*.css` version does not get generated leaving the templates broken.

How about generating the `demo.css` once, storing it somewhere under `styleguide` directory and changing this path to point to it? Once committed, this will make this file always be available to the HTML ones.

styleguide/index.html
```diff
@@ -0,0 +1,96 @@
+<!DOCTYPE html>
+<html lang="en" class="">
+<head>
+
+  <title>NHS - Replacement Publication System</title>
+
+  <meta charset="UTF-8" />
+  <meta name="title" content="NHS - Replacement Publication System" />
+  <meta name="viewport" content="width=device-width, minimum-scale=1.0, maximum-scale=1.0" />
+  <link rel="stylesheet" href="../repository-data/webfiles/src/main/resources/site/css/demo.css" />
```
ziemsky: See comment about `demo.css` against `styleguide/dataset.html` above.

styleguide/publication.html
```diff
@@ -0,0 +1,153 @@
+<!DOCTYPE html>
+<html lang="en" class="">
+<head>
+
+  <title>NHS - Replacement Publication System - Publication</title>
+
+  <meta charset="UTF-8" />
+  <meta name="title" content="NHS - Replacement Publication System" />
+  <meta name="viewport" content="width=device-width, minimum-scale=1.0, maximum-scale=1.0" />
+  <link rel="stylesheet" href="../repository-data/webfiles/src/main/resources/site/css/demo.css" />
```
ziemsky: See comment about `demo.css` against `styleguide/dataset.html` above.

styleguide/series.html
```diff
@@ -0,0 +1,106 @@
+<!DOCTYPE html>
+<html lang="en" class="">
+<head>
+
+  <title>NHS - Replacement Publication System - Series</title>
+
+  <meta charset="UTF-8" />
+  <meta name="title" content="NHS - Replacement Publication System" />
+  <meta name="viewport" content="width=device-width, minimum-scale=1.0, maximum-scale=1.0" />
+  <link rel="stylesheet" href="../repository-data/webfiles/src/main/resources/site/css/demo.css" />
```
ziemsky: See comment about `demo.css` against `styleguide/dataset.html` above.

styleguide/dataset.html
wojtek-oledzki: I would like to avoid duplicated CSS - it opens the door to confusion, and people changing css in `/styleguide` folder directly :(

PR #74, opened 2017-12-11
[RPS-176] Upgrade to latest 12.1.0

wojtek-oledzki:
Updated taxonomy content management - empty publication_taxonomy will be
created when not present. In development mode it will be replaced with
the text content.


repository-data/development/src/main/resources/hcm-actions.yaml
```diff
@@ -0,0 +1,4 @@
+---
+action-lists:
+- 1.1:
+    /content/taxonomies/publication_taxonomy: reload
```
ngta1: Nice

wojtek-oledzki:
@ziemsky the "temporary" bit applies only to the bit where we try to find data in both v2 and the previous value.

Next sprint I will remove `publicationsystem:attachments` and `publicationsystem:Files`.

ziemsky:
Sounds good!

PR #75, opened 2017-12-11
[RPS-204] CI migration: make folder names human-friendly

ziemsky:
EximImport was re-set to its original version and MigratorImporterScript
was added as THE import script to use to import content generated by
the Migrator application; readme has been updated accordingly.

This is to reflect the fact that MigratorImporterScript is tailored
specifically to consume content generated by the Migrator rather than
more generic output of EximExport.

'FOLDER' JSON files generated by the Migrator application are now being
processed by the import script and so the folders are now created with
display names as specified by property 'jcr:localizedName' in files
generated by the Migrator.

It is now also possible to create empty folders upon migration.

### DIFF
--- a/repository-data/application/src/main/resources/hcm-config/configuration/update/EximImport.groovy
+++ b/repository-data/application/src/main/resources/hcm-config/configuration/update/EximImport.groovy
@@ -29,11 +29,11 @@ import org.onehippo.forge.content.exim.core.*
 import org.onehippo.forge.content.exim.core.impl.*
 import org.onehippo.forge.content.exim.core.util.*
 
-class EximImport extends BaseNodeUpdateVisitor {
+class ImportingDocumentVariantInFileUpdateVisitor extends BaseNodeUpdateVisitor {
 
     def fileInJson = true
-    DocumentManager documentManager
-    WorkflowDocumentVariantImportTask importTask
+    def documentManager
+    def importTask
     def sourceBaseFolder
 
     void initialize(Session session) {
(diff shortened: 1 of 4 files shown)


wojtek-oledzki:
Folders!!

Awesomeness

repository-data/development/src/main/resources/hcm-config/main.yaml
```diff
@@ -2,7 +2,7 @@
 definitions:
   config:
     /hippo:configuration/hippo:modules/autoexport/hippo:moduleconfig:
-      autoexport:enabled: true
+      autoexport:enabled: false
```
wojtek-oledzki: revert back to "true" please.

repository-data/application/src/main/resources/hcm-config/configuration/update/EximImport.groovy
```diff
@@ -29,11 +29,11 @@ import org.onehippo.forge.content.exim.core.*
 import org.onehippo.forge.content.exim.core.impl.*
```
wojtek-oledzki: Is this just roll back to original Exim i mporter?

ziemsky: It is, yes.  `MigratorImporterScript` targets files generated by the Migrator application and that means that it no longer qualifies as a generic importer.

Since we're apparently providing vanilla Exim Exporter, it seemed to make sense to continue providing corresponding vanilla Exim Importer as well so when I introduced `MigratorImporterScript`, I've restored the original Exim Importer script.

repository-data/development/src/main/resources/hcm-config/main.yaml
ziemsky: Done.

repository-data/application/src/main/resources/hcm-config/configuration/update/EximImport.groovy
wojtek-oledzki: :+1:

PR #78, opened 2017-12-13
[RPS-206] Add resource links to datasets

jacsamell:
Added resource links to datasets in the same format as the publication.
Also added migration for resource links from Nesstar data.


acceptance-tests/src/test/resources/features/site/datasets.feature
```diff
@@ -14,8 +14,11 @@ Scenario: Search individual data set by name
 
 Scenario: Download individual files from data set
```
ziemsky: Scenario name could probably be updated to reflect the fact that it's not just about the files anymore.

jacsamell: 👍

PR #80, opened 2017-12-14
[RPS-203] Protect dataset properties

ngta1:
Protect dataset getters from illegal access when the parent publication
flag is set to upcoming.

### DIFF
--- a/site/src/main/java/uk/nhs/digital/ps/beans/Dataset.java
+++ b/site/src/main/java/uk/nhs/digital/ps/beans/Dataset.java
@@ -5,40 +5,53 @@ import org.onehippo.cms7.essentials.dashboard.annotations.HippoEssentialsGenerat
 import org.hippoecm.hst.content.beans.Node;
 import java.util.Calendar;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


site/src/main/java/uk/nhs/digital/ps/beans/Dataset.java
```diff
@@ -95,4 +109,42 @@ public Publication getParentPublication() {
...
+                || propertiesPermittedWhenUpcoming.contains(propertyKey);
+
+        if (!isPropertyPermitted) {
+            throw new DataRestrictionViolationException(
+                "Property is not available when parent publication is flagged as 'not publicly accessible': " + propertyKey
+            );
+        }
+    }
+
+    public boolean isParentPublicationPubliclyAccessible() {
```
wojtek-oledzki: I wouldn't reveal implementation detail in the function name, `isPubliclyAccessible()` is absolutely enough in my opinion.

ngta1: Fair point, will amend.

PR #82, opened 2017-12-15
[RPS-201] Fix preview

wojtek-oledzki:
This involves changes to CMS distribution package.

Also:
* fixed migration for Mac OS
* moved non-prod vhosts to `development` module

### DIFF
--- a/migrator/pom.xml
+++ b/migrator/pom.xml
@@ -86,6 +86,12 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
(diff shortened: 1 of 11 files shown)


migrator/pom.xml
```diff
@@ -86,6 +86,12 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <groupId>org.apache.tika</groupId>
```
ngta1: ❤️ 🍏 👍

PR #90, opened 2017-12-20
[RPS-220] Set Freemarker template MIME type

ngta1:
FreeMarker recommend using ftlh and ftlx file extensions to turn on HTML
and XML auto-escaping (along with a "incompatible_improvements" setting
which was introduced in version 2.3.24). All template files renamed to
use the .ftlh extension. Removed legacy ?html escaping from templates.

To check the status of FreeMarker configuration, add the following to
a  template:
<p>Output format: ${.output_format}
<p>Auto-escaping: ${.auto_esc?c}


repository-data/webfiles/src/main/resources/site/freemarker/publicationsystem/dataset.ftlh
```diff
@@ -1,11 +1,14 @@
...
+<#include "./macro/structured-text.ftlh">
 <@hst.setBundle basename="publicationsystem.headers"/>
 
 <#assign dateFormat="dd/MM/yyyy"/>
 <#assign formatFileSize="uk.nhs.digital.ps.directives.FileSizeFormatterDirective"?new() >
 <#assign formatCoverageDates="uk.nhs.digital.ps.directives.CoverageDatesFormatterDirective"?new() >
 
+<p>Output format: ${.output_format}
+<p>Auto-escaping: ${.auto_esc?c}
+
```
jacsamell: Is this just for testing?

ngta1: yes, forgot to remove before commit :-).  Will do that now...

PR #93, opened 2017-12-21
[RPS-207] Migration Report

jacsamell:
Added a migration report that will capture errors we have migrating
the Nesstar data. This now enables the migration to continue if
specific datasets have errors.

Also added Unique ID (P-code) to the end of the dataset summary.

### DIFF
--- a/migrator/src/main/java/uk/nhs/digital/ps/migrator/model/hippo/Attachment.java
+++ b/migrator/src/main/java/uk/nhs/digital/ps/migrator/model/hippo/Attachment.java
@@ -3,6 +3,8 @@ package uk.nhs.digital.ps.migrator.model.hippo;
 import org.apache.tika.Tika;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import uk.nhs.digital.ps.migrator.MigrationReport;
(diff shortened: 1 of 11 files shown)


migrator/src/main/java/uk/nhs/digital/ps/migrator/model/hippo/Attachment.java
```diff
@@ -88,8 +89,13 @@ public void download() {
                     log.debug("Downloaded {} Mb / {} Mb", getMb(bytesRead), getMb(bytesExpected));
                 }
             }
-        } catch (IOException e) {
-            throw new RuntimeException("Error occurred downloading attachment", e);
+            return true;
+        } catch (Exception e) {
+            file.delete(); // Attempt to delete the file as it may be partially downloaded
```
ngta1: Probably unlikely to happen, but could the partially downloaded file be in use/locked by a process, and an exception thrown by the delete attempt?

jacsamell: Good point, If this happens the migration will just fail so hopefully it is unlikely enough that a total failure is ok as we can resolve the issue and run again.

PR #96, opened 2017-12-27
[RPS-232] Faceted search month bug fix

jacsamell:
Fixed an issue with the display of the month name in faceted search.


repository-data/webfiles/src/main/resources/site/freemarker/common/facets.ftlh
```diff
@@ -1,6 +1,5 @@
 <#include "../include/imports.ftlh">
-<@hst.setBundle basename="month-names"/>
-<@hst.setBundle basename="document-types"/>
+<@hst.setBundle basename="document-types,month-names"/>
```
ivorc: Suggest adding a comment to make it clear that we need to do a single setBundle only

jacsamell: Good idea

PR #97, opened 2017-12-27
[RPS-216] Page not found page

jacsamell:
Changed the page not found page so it uses the base layout so the
styling is the same as the rest of the site. Also set the HTTP status
code to 404.

Added to the sitemap so if you enter an invalid URL you will be
directed to the new 404 page.

Also cleaned up some items in the sitemap that either weren't used
or were specified in the wrong place.


acceptance-tests/src/test/resources/features/site/errors.feature
```diff
@@ -0,0 +1,14 @@
+Feature: As a site user I should be shown a styled error page
+    when an error occurs on the site.
+
+    Scenario: Navigate to an invalid root URL
+        Given I navigate to the "Invalid Root" page
```
jacsamell: Ivor: suggest not capitalising page names

PR #98, opened 2017-12-28
[RPS-210] Add terms and conditions

ngta1:
Added new document type called "About" which will cater for general site
documents such as terms and conditions, accessibility, privacy policy etc. Added
footer component to basepage, along with initial link to terms and conditions document.

Enabled a populated version of terms and conditions document to be bootstraped in the
development module, with a blank document being bootstraped in other environments on
first deployment for the user to populate and maintain (similar to taxonomy). For
subsequent deployments, as the document will already exist, the user maintained version
will not be overwritten.

Note: new document type currently resides under the ps namespace.  Separate ticket to
move this into the common namespace.


ngta1:
Thanks for feedback.  Changes now applied.

PR #99, opened 2017-12-28
[RPS-239] Added validation for dataset creation

jacsamell:
Extended the logic for testing publication validation so it can be
applied to datasets as well.

This has involved fixing some of the config for datasets as well.


acceptance-tests/src/test/java/uk/nhs/digital/ps/test/acceptance/steps/TestDataSteps.java
```diff
@@ -43,10 +44,18 @@ public void clearTestData() {
      */
     @After(value = "@DiscardAfter", order = 500)
     public void discardEditedPublication() throws Throwable {
```
ngta1: Could rename to "discardEditedDocument" to keep consistent with other publication to document changes?

jacsamell: 👍

PR #102, opened 2017-12-29
[RPS-227] Enforce HTTPS connections

jacsamell:
Set HTTPS to be the default scheme (default was HTTP).
If you visit the http URL you will be redirected to https.
Overwritten this value for AWS and localhost as these envirnoments
don't have certificates set up, therefore this will only affect
OnDemand.

### DIFF
--- a/repository-data/application/src/main/resources/hcm-config/hst/hosts.yaml
+++ b/repository-data/application/src/main/resources/hcm-config/hst/hosts.yaml
@@ -34,6 +34,7 @@ definitions:
                 jcr:primaryType: hst:virtualhost
               jcr:primaryType: hst:virtualhost
             jcr:primaryType: hst:virtualhost
+          hst:scheme: http
           hst:showcontextpath: false
           jcr:primaryType: hst:virtualhost
         hst:cmslocation: http://blackhole-hippo-authoring.web.dev.ps.nhsd.io/cms
@@ -70,6 +71,7 @@ definitions:
                 jcr:primaryType: hst:virtualhost
               jcr:primaryType: hst:virtualhost
             jcr:primaryType: hst:virtualhost
+          hst:scheme: http
           hst:showcontextpath: false
           jcr:primaryType: hst:virtualhost
         hst:cmslocation: http://hippo-authoring.web.dev.ps.nhsd.io/cms
@@ -135,4 +137,5 @@ definitions:
           jcr:primaryType: hst:virtualhost
         hst:cmslocation: https://cms-nhs.test.onehippo.com
         jcr:primaryType: hst:virtualhostgroup
+      hst:scheme: https
       jcr:primaryType: hst:virtualhosts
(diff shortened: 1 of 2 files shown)


repository-data/application/src/main/resources/hcm-config/hst/hosts.yaml
```diff
@@ -34,6 +34,7 @@ definitions:
                 jcr:primaryType: hst:virtualhost
               jcr:primaryType: hst:virtualhost
             jcr:primaryType: hst:virtualhost
+          hst:scheme: http
```
ivorc: Is it worth adding a comment: # AWS environment will be decommissioned soon, so we won't invest the effort in setting up SSL

repository-data/application/src/main/resources/hcm-config/hst/hosts.yaml
```diff
@@ -70,6 +71,7 @@ definitions:
...
```
ivorc: Is it worth adding a comment: # No SSL in local dev environment for simplicity

repository-data/development/src/main/resources/hcm-config/hst/hosts.yaml
```diff
@@ -19,6 +19,7 @@ definitions:
             hst:homepage: root
             hst:mountpoint: /hst:hst/hst:sites/common
             jcr:primaryType: hst:mount
+          hst:scheme: http
```
ivorc: Is it worth adding a comment: # No SSL in local dev environment for simplicity

repository-data/development/src/main/resources/hcm-config/hst/hosts.yaml
```diff
@@ -41,6 +42,7 @@ definitions:
               hst:mountpoint: /hst:hst/hst:sites/common
               jcr:primaryType: hst:mount
             jcr:primaryType: hst:virtualhost
+          hst:scheme: http
```
ivorc: Is it worth adding a comment: # No SSL in local dev environment for simplicity

repository-data/application/src/main/resources/hcm-config/hst/hosts.yaml
jacsamell: Agree this would be ideal but unfortunately we can't have comments in YAML at the moment because the YAML Formatter strips them out! 😞

repository-data/development/src/main/resources/hcm-config/hst/hosts.yaml
jacsamell: Can't comment in YAML, see above

repository-data/development/src/main/resources/hcm-config/hst/hosts.yaml
jacsamell: Can't comment in YAML, see above

repository-data/application/src/main/resources/hcm-config/hst/hosts.yaml
jacsamell: Can't comment in YAML, see above

PR #103, opened 2017-12-29
[RPS-240] Series validation tests

jacsamell:
Added validation tests for series.
Also spotted that the other documents were missing tests for mandatory
title and summary so added these as well.

Made some improvements to CMS tests in general so they should fail
a lot less often when there isn't actually a problem.


acceptance-tests/src/test/java/uk/nhs/digital/ps/test/acceptance/pages/ContentPage.java
```diff
@@ -281,7 +291,18 @@ private WebElement findTakeOffline() {
...
 
+    /**
+     * Sometimes when send a click, it just highlights the button but doesn't click it..
+     * The second time always seems to work so only try twice
+     */
     private void clickButtonOnModalDialog(String buttonText) {
+        try {
+            clickButtonOnModalDialogOnce(buttonText);
+        } catch (TimeoutException e) {
+            clickButtonOnModalDialogOnce(buttonText);
```
ngta1: Nice find 👍

PR #110, opened 2018-01-03
[RPS-255] Accessibility WCAG2 AA Compliance

jacsamell:


repository-data/webfiles/src/main/resources/site/freemarker/publicationsystem/publications-overview.ftl
```diff
@@ -13,11 +13,11 @@
 
 
 <#-- CONTENT: -->
-<section class="document-content">
+<section class="document-content" aria-label="Publication Overview">
```
ziemsky: I'm definitely not the right person to lecture native speakers on English, but wouldn't this be more accurate if it said 'Publication**s** Overview'?

jacsamell: Good spot! Will update 🙂

ziemsky: Cool! Feel free to merge straight after, I don't need to see the fix :wink:

PR #116, opened 2018-01-05
[RPS-249] Update config to prepare to send logs to splunk

ngta1:
Basic demo of exporting logs to Splunk.  Splunk Enterprise
set up locally with temp developer licence, and a Http Event
Collector setup to receive logs on that instance.
Tested locally. Further ticket to configure to point to
NHS Splunk instance with dashboard setup etc.


conf/log4j2-dev.xml
```diff
@@ -1,136 +1,186 @@
...
+
+        <!-- splunk CMS -->
+        <Http name="splunk-cms"
+            url="${sys:splunk.url}"
+            token="${sys:splunk.token}"
+            batch_size_count="${sys:splunk.batch.size.count}"
+            disableCertificateValidation="true"
+            source="nhsdhippocms"
+            sourcetype="hippo-cms">
+            <PatternLayout pattern="%m"/>
```
wojtek-oledzki: @ngta1 can you test the "<LookupFilter ...>" tag. This might simplify the config. Check the hippo-cms and hippo-site files appenders above.

pom.xml
```diff
@@ -74,6 +74,10 @@
         <!-- revision variable is set by automation tools. dev-SNAPSHOT is the
         default, local value. Please don't change it. -->
         <revision>dev-SNAPSHOT</revision>
+
+        <splunk.token></splunk.token>
```
wojtek-oledzki: a short comment will help fellow developers to understand what is it for :)

site/pom.xml
```diff
@@ -125,6 +125,11 @@
             <artifactId>junit-dataprovider</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
wojtek-oledzki: do we need that here?

site/pom.xml
```diff
@@ -125,6 +125,11 @@
...
+            <groupId>com.splunk.logging</groupId>
+            <artifactId>splunk-library-javalogging</artifactId>
+            <version>1.5.2</version>
```
ziemsky: Seeing that the `splunk-library-javalogging` has already been defined in the parent POM's `<dependencyManagement>` section (:+1: ), the declaration in this POM file doesn't need `<version>` specified. 

Te idea behind `<dependencyManagement>` is that it defines a pool of common 'blessed' dependencies for child POMs to choose from in which case they only need to specify `groupId` and `artifactId` themselves.

site/pom.xml
ziemsky: If you mean _'do we need this declaration if we have it declared in parent POM already'_, @wojtek-oledzki,  then see comment below about `dependencyManagement` below.

If you're asking whether we need this dependency in Site module at all - I'll defer to @ngta1 :wink:

pom.xml
ngta1: Yep, comments added.

site/pom.xml
ngta1: Thanks, version removed.

site/pom.xml
ngta1: Good question.  Even though it is defined in the parent POM, I tried removing this dependency from the site pom, but the logging to Splunk stops for some reason (for site, cms and audit).  I have however removed the version as per @ziemsky comment below.

pom.xml
```diff
@@ -74,6 +74,15 @@
         <!-- revision variable is set by automation tools. dev-SNAPSHOT is the
         default, local value. Please don't change it. -->
         <revision>dev-SNAPSHOT</revision>
+
+        <!-- Unique token for the for the Splunk HTTP Event Collector (HEC)
```
woledzki: small typo
> for the for the

PR #117, opened 2018-01-05
[RPS-269] Tidied up acceptance tests

jacsamell:
Separated the test data from the test factory and the actual tests
for series and datasets.

### DIFF
--- a/acceptance-tests/src/test/java/uk/nhs/digital/ps/test/acceptance/models/DatasetBuilder.java
+++ b/acceptance-tests/src/test/java/uk/nhs/digital/ps/test/acceptance/models/DatasetBuilder.java
@@ -1,0 +1,76 @@
+package uk.nhs.digital.ps.test.acceptance.models;
+
+import java.time.Instant;
+
+public class DatasetBuilder {
+
+    private String name;
+    private String title;
+    private String summary;
+    private Instant nominalDate;
+
+    public static DatasetBuilder newDataset() {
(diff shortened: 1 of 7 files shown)


acceptance-tests/src/test/java/uk/nhs/digital/ps/test/acceptance/models/DatasetBuilder.java
```diff
@@ -0,0 +1,54 @@
+package uk.nhs.digital.ps.test.acceptance.models;
+
+import java.time.Instant;
+
+public class DatasetBuilder {
```
ziemsky: Other builders in this package are immutable; they follow approach where all the `with*` methods return new instance of a builder. This is achieved through the use of `PropertySetter` interface, `cloneAndAmend(final PropertySetter propertySetter)` and copy constructors (+ the default, parameterless consturctor made private).

To keep things consistent, would you mind bringing this class in line with the other builders? `PublicationSeriesBuilder` should provide sufficient example.

jacsamell: Good idea, I'll update the builder. Tried to use intellijs builder generator but it creates in this format so I'll just create it manually 🙂

ziemsky: :+1:

PR #130, opened 2018-01-11
[RPS-259] Removed hardcoded user accounts

jacsamell:
Removed all the hard coded user accounts (including hashed passwords)
except for admin. We are leaving admin with a random hashed password
as this is better than having admin:admin and we will delete the
admin account as soon as we have set up individual accounts in prod.


wojtek-oledzki:
have a look at `group.base.yaml` as well

jacsamell:
Good spot. I'll fix in the morning

PR #134, opened 2018-01-12
[RPS-258] Prepare content for search improvements

wojtek-oledzki:
This commit adds "common:searchable" flag to all documents. It's set to
"true" be default and autocreated on all documents.

Following commit will add new functionality to remove selected files
from search results - like upcomming publications and datasets.


repository-data/application/src/main/resources/hcm-config/configuration/update/Migration20180112Searchable.groovy
```diff
@@ -0,0 +1,63 @@
+package org.hippoecm.frontend.plugins.cms.admin.updater
+
+import javax.jcr.*
+import org.onehippo.repository.update.BaseNodeUpdateVisitor
+
+/**
+ * This migration migrate "hippo:resource" type to "publicationsystem:resource".
+ * "publication:resource" type has "hippo:text" property always set to empty String
+ * which make PDF attachments NOT searchable.
```
jacsamell: Don't think that comment is accurate ;)

PR #136, opened 2018-01-15
[RPS-296] CI migration: modify links filtering

ziemsky:
From now on all links with marker texts are filtered out, regardless
of whether they belong to Compendium or not, but only those that are
from outside Compendium are reported on.

All excluded links and attachments are logged.

### DIFF
--- a/migrator/src/main/java/uk/nhs/digital/ps/migrator/task/ImportableItemsFactory.java
+++ b/migrator/src/main/java/uk/nhs/digital/ps/migrator/task/ImportableItemsFactory.java
@@ -1,6 +1,6 @@
 package uk.nhs.digital.ps.migrator.task;
 
-import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
 import uk.nhs.digital.ps.migrator.config.ExecutionParameters;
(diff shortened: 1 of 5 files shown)


migrator/src/main/java/uk/nhs/digital/ps/migrator/task/ImportableItemsFactory.java
```diff
@@ -62,19 +66,20 @@ public Folder toFolder(final Folder parentFolder, final Catalog catalog) {
...
 
             List<NesstarResource> resources = exportedPublishingPackage.getResources();
-            // compendium migration do not require PDFs to be migrated over.
-            List<Attachment> attachments = getAttachments(exportedPublishingPackage).stream()
-                                            .filter(a -> !(a.getMimeType().equalsIgnoreCase("application/pdf")
-                                                && parentFolder.getJcrPath().contains("compendium-of-population-health-indicators")))
-                                            .collect(Collectors.toList());
+
+            List<Attachment> attachments = getAttachments(exportedPublishingPackage);
+            attachments = filterAttachments(pCode, attachments, isWithinCompendiumStructure(parentFolder));
```
ngta1: 👍

PR #142, opened 2018-01-16
[RPS-300] Removed 'nominal' from the description of publication date

jacsamell:
Fixed some of the formatting of the headers.
Also fixed and ran maven dependency updater.


wojtek-oledzki:
Why are we not using `<@fmt.message key="headers.date_rang"/>` etc.?

jacsamell:
Updated to use headers file

PR #145, opened 2018-01-17
[RPS-292] Indicate document type on content pages

ngta1:
Added document type label on publication, series and dataset pages

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# NHS Digital Publication System - Hippo CMS
+# NHS Digital Website - Hippo CMS
 
 [![Build Status][build-status]][travis-project-page]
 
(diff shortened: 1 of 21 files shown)


repository-data/application/src/main/resources/hcm-content/content/documents/administration/publication-system/headers.yaml
```diff
@@ -133,6 +154,9 @@
     - headers.geographical-granularity
     - headers.publication-date
     - headers.next-publication-date
+    - headers.publication-label
```
wojtek-oledzki: maybe should be in a separate file - `administration/publication-system/labels.yaml`

ngta1:
Labels now added to separate file, so separate from headers.

PR #146, opened 2018-01-19
[RPS-291] Indicate document type of each search result

ngta1:
Also includes refactoring of search result templates to use a separate
macro for each document type.  Fixed issue where dataset nominal
date was not being displayed on search results page. Includes small
change to password rules acceptance test to try resolve random fails.
Update to Makefile to replace publication-system with website.

### DIFF
--- a/repository-data/webfiles/src/main/resources/site/freemarker/common/macro/results.ftl
+++ b/repository-data/webfiles/src/main/resources/site/freemarker/common/macro/results.ftl
@@ -1,0 +1,54 @@
+<#ftl output_format="HTML">
+<@hst.setBundle basename="publicationsystem.labels"/>
+<#assign formatRestrictableDate="uk.nhs.digital.ps.directives.RestrictableDateFormatterDirective"?new() />
+
+<#macro searchResutls items>
+    <#list items as document>
+        <#if document.class.name == "uk.nhs.digital.ps.beans.Publication">
+            <@publication item=document />
+        <#elseif document.class.name == "uk.nhs.digital.ps.beans.Series">
+            <@series item=document />
+        <#elseif document.class.name == "uk.nhs.digital.ps.beans.Dataset">
+            <@dataset item=document />
+        </#if>
+    </#list>
+</#macro>
+
+<#macro publication item>
+    <div class="push-double--bottom" data-uipath="ps.search-results.result">
+        <h3 class="flush zeta" data-uipath="ps.search-results.result.type" style="font-weight:bold"><@fmt.message key="labels.publication"/></h3>
+        <p class="flush">
+            <a href="<@hst.link hippobean=item.selfLinkBean/>" title="${item.title}" data-uipath="ps.search-results.result.title">
+                ${item.title}
+            </a>
+        </p>
+        <p class="flush zeta" data-uipath="ps.search-results.result.date"><@formatRestrictableDate value=item.nominalPublicationDate/></p>
+        <p class="flush" data-uipath="ps.search-results.result.summary"><@truncate text=item.summary.firstParagraph size="300"/></p>
+    </div>
+</#macro>
+
+<#macro series item>
+    <div class="push-double--bottom" data-uipath="ps.search-results.result">
+        <h3 class="flush zeta" data-uipath="ps.search-results.result.type" style="font-weight:bold"><@fmt.message key="labels.series"/></h3>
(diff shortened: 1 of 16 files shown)


wojtek-oledzki:
Looks good, and thanks for the Makefile fix

repository-data/webfiles/src/main/resources/site/freemarker/common/searchresults.ftl
```diff
@@ -11,17 +10,7 @@
...
-                    <p class="flush">
-                        <a href="<@hst.link hippobean=publication.selfLinkBean/>" title="${publication.title}" data-uipath="ps.search-results.result.title">
-                            ${publication.title}
-                        </a>
-                    </p>
-                    <p class="flush zeta" data-uipath="ps.search-results.result.date"><@formatRestrictableDate value=publication.nominalPublicationDate/></p>
-                    <p class="flush" data-uipath="ps.search-results.result.summary"><@truncate text=publication.summary.firstParagraph size="300"/></p>
-                </div>
-            </#list>
+            <#include "./macro/results.ftl">
```
ziemsky: Hmm, I'd suggest that:
* the `<#list...` block in the `results.ftl` be wrapped in a macro
* this `<#include...` moved to the top of the file
* make this line call the macro rather than include.

I see 'includes' more as a declaration of some re-usable functionality rather than the application of it. To my mind, this is similar to Java imports or C headers' inclusions - at the beginning of the file we declare what we will be used and later in the body we actually use it by calling individual macros defined in the included files.

Also, having all reusable content wrapped in macros makes it easier to pick-and-choose individual macros from the included file; conversly, having a bit of inlined content (i.e. not wrapped in a macro) in the included file forces us to use it as soon as we include the file, which may or may not be applicable in all situations.

I emphasise that this is just my subjective view so feel free to ignore it - but I think this keeps the files a bit more 'predictable' and easier to read and reusable.

repository-data/webfiles/src/main/resources/site/freemarker/common/macro/results.ftl
```diff
@@ -0,0 +1,53 @@
...
+
+<#list pageable.items as document>
+    <#if document.class.name == "uk.nhs.digital.ps.beans.Publication">
+        <@publication item=document />
+    <#elseif document.class.name == "uk.nhs.digital.ps.beans.Series">
+        <@series item=document />
+    <#elseif document.class.name == "uk.nhs.digital.ps.beans.Dataset">
+        <@dataset item=document />
+    </#if>
+</#list>
```
ziemsky: See my comment  in `searchresults.ftl` about wrapping this block in a macro.

site/src/main/java/uk/nhs/digital/ps/beans/BaseDocument.java
```diff
@@ -39,4 +44,22 @@ public HippoBean getSelfLinkBean() {
     protected void assertPropertyPermitted(String propertyName) {
         // To be overwritten by subclasses for specific implementation
     }
+
+    /**
+     * Converts given {@linkplain Calendar} to {@linkplain RestrictableDate}.
+     */
+    protected RestrictableDate nominalPublicationDateCalendarToRestrictedDate(final Calendar calendar) {
```
ziemsky: Hmm, we seem to be adding more methods to `BaseDocument` that not all its extending classes use - this applies to this and all the `assertPropertypermitted`-related methods. For example, `Series` doesn't use any of them and yet it inherits - and exposes - them.

I wonder if this is the time where we should consider applying composition over inheritance and move this and `assertPropertypermitted`-related methods to an external class, where `Publication` and `Dataset` would call/wrap them rather than inherit them from the base class. While the `assertPropertypermitted` & Co. are arguably more generic in nature, the `nominalPublicationDateCalendarToRestrictedDate` is much more specific.

I worry that if we continue this trend, at some point we will make BaseDocument so bloated, and it's responsibilities so blurred, that we'll find it hard to understand and maintain. 

BTW, it's my fault that we've started down this path as, if I remember correctly, it was me who advised Jake to stick the `assertPropertypermitted`-related methods here :wink:

Happy to talk about it more if you want and if you think there is enough merit in the above comment.

wojtek-oledzki: something to consider for next PR :D

PR #147, opened 2018-01-23
[RPS-282] Document taxonomy tree

jacsamell:
Added an event bus listener that will store all the branches in the
taxonomy hierarchy above the leaves selected for the document.

### DIFF
--- a/cms/src/main/java/uk/nhs/digital/ps/modules/FullTaxonomyModule.java
+++ b/cms/src/main/java/uk/nhs/digital/ps/modules/FullTaxonomyModule.java
@@ -1,0 +1,153 @@
+package uk.nhs.digital.ps.modules;
+
+import org.apache.jackrabbit.value.StringValue;
+import org.onehippo.repository.events.HippoWorkflowEvent;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
+import javax.jcr.Node;
+import javax.jcr.RepositoryException;
+import javax.jcr.Session;
+import javax.jcr.Value;
+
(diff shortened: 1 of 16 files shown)


cms/src/main/java/uk/nhs/digital/ps/modules/FullTaxonomyModule.java
```diff
@@ -0,0 +1,153 @@
...
+    private void createFullTaxonomyProperty(Node document) {
+        Set<String> fullTaxonomyKeys = getFullTaxonomyKeys(document);
+        Value[] fullTaxonomy = fullTaxonomyKeys.stream()
+            .map(StringValue::new)
+            .toArray(Value[]::new);
+
+        try {
+            document.setProperty(FULL_TAXONOMY_PROPERTY, fullTaxonomy);
+        } catch (RepositoryException re) {
+            log.error("Failed to set full taxonomy on node: " + document);
```
ngta1: Should the exception details be passed as well as the document?  Does it need to be thrown?

cms/src/main/java/uk/nhs/digital/ps/modules/FullTaxonomyModule.java
```diff
@@ -0,0 +1,153 @@
...
+        return getRepoValue(() -> node.getProperty("hippotaxonomy:key").getString());
+    }
+
+    private Stream<Node> getDescendantTaxonomyNodes(Node node) {
+        return streamChildTaxonomyNodes(node)
+            .flatMap(childNode -> Stream.concat(
+                Stream.of(childNode),
+                getDescendantTaxonomyNodes(childNode)
+            ));
+    }
```
ngta1: 👍 , 👍  and 👍 !

cms/src/main/java/uk/nhs/digital/ps/modules/FullTaxonomyModule.java
jacsamell: Good spot!
Don't want to throw as we are quite low down and don't want to interupt the triggering event but should definitley be logging the exception, have updated. 🙂

PR #149, opened 2018-01-23
[RPS-319] Sticky faceted search

jacsamell:
The selected facets will persist when a new search query is entered


site/src/main/java/uk/nhs/digital/common/components/SearchComponent.java
```diff
@@ -132,6 +136,8 @@ private HstQuery buildQuery(HstRequest request) {
                     constraint(".").contains(queryIncWildcards)
                 )
             ))
+            .limit(10)
+            .offset(getCurrentPage(request) * 10)
```
ngta1: Is the "10" the same pageSize configurable value in (search.yaml - /hst:hst/hst:configurations/common/hst:pages/search/main) or is it used for something else?  If same, the SearchComponent uses paramInfo.getPageSize() to retrieve this value, so could use this instead?

jacsamell: Have removed this as discussed with @wojtek-oledzki

PR #157, opened 2018-01-29
[RPS-313] Added migration of mapped fields

jacsamell:
Read in the new mapping spreadsheet and map the fields as part
of the CI migration.

Also added deleting of documents in the cms tests so the data
is fully cleaned up after running any test.

### DIFF
--- a/migrator/src/main/java/uk/nhs/digital/ps/migrator/config/ExecutionParameters.java
+++ b/migrator/src/main/java/uk/nhs/digital/ps/migrator/config/ExecutionParameters.java
@@ -1,12 +1,12 @@
 package uk.nhs.digital.ps.migrator.config;
 
+import static java.util.Arrays.asList;
+import static uk.nhs.digital.ps.migrator.misc.Descriptor.describe;
+
 import uk.nhs.digital.ps.migrator.misc.Descriptor;
 
 import java.nio.file.Path;
 import java.util.List;
-
-import static java.util.Arrays.asList;
-import static uk.nhs.digital.ps.migrator.misc.Descriptor.describe;
 
 public class ExecutionParameters {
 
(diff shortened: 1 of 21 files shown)


migrator/src/main/java/uk/nhs/digital/ps/migrator/config/ExecutionParameters.java
```diff
@@ -103,6 +104,14 @@ public void setTaxonomyMappingImportPath(Path taxonomyMappingImportPath) {
         this.taxonomyMappingImportPath = taxonomyMappingImportPath;
     }
 
+    public Path getFieldMappingImportPath() {
+        return fieldMappingImportPath;
```
ziemsky: Could be worth emphasising that this is CI-specific mapping. Other properties use prefix  'nesstar' for this purpose, so how about renaming this property and the command line arg to, say, `nesstarFieldMappingImportPath`?

Both for consistency and to indicate that this should only be used for this specific type of mapping file (so as to not to suggest that this is for some generic/universal type of mapping).

migrator/src/main/java/uk/nhs/digital/ps/migrator/report/IncidentType.java
```diff
@@ -151,6 +151,47 @@
...
+            "Datasets get certain fields applied as defined in the mapping file"
+            + "\nfor each Dataset but the provided mapping file"
+            + "\n was missing mapping for a certian field for a given dataset."
+            + "\n"
+            + "\nUpdated mapping file should be provided, one that includes"
+            + "\nthe missing field."
+    ),
+    FIELD_MAPPING_INVALID(
+        FIELD_NOT_MIGRATED,
+        "Field mapping invalid field for Dataset",
```
ziemsky: _Field mapping is invalid for Dataset_?

jacsamell:
Good suggestions, have updated.

PR #161, opened 2018-01-30
[RPS-308] Better import mechanism

ziemsky:
Fixed the mistake of overwriting the main readme with content of the
Migrator's readme.

### DIFF
--- a/migrator/README.md
+++ b/migrator/README.md
@@ -4,75 +4,198 @@ The new, Hippo CMS based publication system requires that data from legacy syste
 available from the go-live.
 
 At the moment of writing, there are two such systems - NHS Digital [Clinical Indicators portal] and [GOSS]-based legacy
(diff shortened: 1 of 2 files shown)


migrator/README.md
```diff
@@ -9,69 +9,192 @@ Publications System.
...
 before the next file is processed. 
 
-Click `Execute` and watch the log.       
+Click `Execute` to trigger the import. At the end examine the log presented in the [Updater Editor] and application log
+file as not all exceptions bubble up to the editor's log.    
+
+## Server side activities (local and remote environments)
+
+### Exporting via XML Export
+<a name="exportingViaXmlExport"></a>
```
wojtek-oledzki: :D we don't need that :) Just use `[bla bla](#exporting-via-xml-export)`

PR #164, opened 2018-01-31
[RPS-357] Removed taxonomy from doc site pages

jacsamell:


jacsamell:
Good shout @ngta1, have done so.

PR #173, opened 2018-02-01
[RPS-350] Made Doc type facets consistent

jacsamell:
Made them use the configurable doc type labels.
Also fixed a bug where taxonomy terms would repeat if the
taxonomy key was invalid.

### DIFF
--- a/repository-data/application/src/main/resources/hcm-content/content/documents/administration/document-types.yaml
+++ b/repository-data/application/src/main/resources/hcm-content/content/documents/administration/document-types.yaml
@@ -22,10 +22,10 @@
     - publicationsystem:dataset
     - publicationsystem:publication
     resourcebundle:messages:
-    - Archive
-    - Series
-    - Data Set
-    - Statistical Publication
+    - labels.archive
+    - labels.series
+    - labels.dataset
+    - labels.publication
   /document-types[2]:
     hippo:availability:
     - preview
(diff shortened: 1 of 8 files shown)


site/src/main/java/uk/nhs/digital/common/components/FacetComponent.java
```diff
@@ -3,20 +3,45 @@
...
+
+        private TaxonomyWrapper(Taxonomy taxonomy) {
+            this.taxonomy = taxonomy;
+        }
+
+        public String getValueName(String key) {
+            Category taxonomyCategory = taxonomy.getCategoryByKey(key);
+            if (taxonomyCategory == null) {
+                log.error("No taxonomy for key: " + key);
+                return "Unknown Option: " + key;
```
wojtek-oledzki: Something to check with PO. 
In terms of presentation, maybe it's better to return nothing, or do some guessing like replace "-" with " " etc...

repository-data/application/src/main/resources/hcm-content/content/documents/administration/document-types.yaml
```diff
@@ -70,9 +70,9 @@
     - publicationsystem:dataset
     - publicationsystem:publication
     resourcebundle:messages:
-    - Series
-    - Data Set
-    - Statistical Publication
+    - labels.series
+    - labels.dataset
+    - labels.publication
```
wojtek-oledzki: can you add "Archive" as well pls

wojtek-oledzki: oh wait, it's not merged yet :D, never mind.

site/src/main/java/uk/nhs/digital/common/components/FacetComponent.java
jacsamell: Have updated after input from Karen, happy for it not be be pretty as it should only happen with bad data on test/uat and shouldn't ever happen on prod.

PR #178, opened 2018-02-06
[DW-0000] Adding binaries import and docbase cleaner scripts

giacomolm:
Those scripts are needed to import binaries and convert the docbase path to UUID

### DIFF
--- a/repository-data/application/src/main/resources/hcm-config/configuration/update/clean-docbase-values-having-paths.groovy
+++ b/repository-data/application/src/main/resources/hcm-config/configuration/update/clean-docbase-values-having-paths.groovy
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2015-2017 Hippo B.V. (http://www.onehippo.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


repository-data/application/src/main/resources/hcm-content/configuration/update/registry-1.yaml
```diff
@@ -0,0 +1,30 @@
+---
+/hippo:configuration/hippo:update/hippo:registry/Clean Hippo Mirror Docbase Values Having Paths:
+  hipposys:batchsize: 10
+  hipposys:description: A example script to clean interim hippo:docbase values of
+    Hippo Mirror nodes having paths instead of UUIDs.
+  hipposys:dryrun: false
+  hipposys:parameters: '{}'
+  hipposys:query: //element(*)[jcr:like(@hippo:docbase,'/content/%')]
+  hipposys:script: "/*\r\n * Copyright 2015-2017 Hippo B.V. (http://www.onehippo.com)\r\
```
wojtek-oledzki: can you add it the same way the other scripts are added - makes it easy to ready

like
https://github.com/NHS-digital-website/hippo/blob/master/repository-data/application/src/main/resources/hcm-config/configuration/update/MigratorImporterScript.yaml#L24-L26

https://github.com/NHS-digital-website/hippo/blob/master/repository-data/application/src/main/resources/hcm-config/configuration/update/MigratorImporterScript.groovy

giacomolm: Good point! Thanks

giacomolm:
Changes done

PR #184, opened 2018-02-07
[RPS-337] CI landing pages add repeating sections

ngta1:
Add new compound type to allow for repeating text/resources sections
on CI landing page. This will allow constuction of content to fit
in with requirements for the SHMI landing page.  Within this repeating
section is a Rich Text Editor field, which is restricted to only allow
basic formatting (bold, italic, underline) and numbered or bullet
point lists.


repository-data/application/src/main/resources/hcm-config/namespaces/hippostd/html.yaml
```diff
@@ -0,0 +1,5 @@
+---
+definitions:
+  config:
+    /hippo:namespaces/hippostd/html/editor:templates/_default_:
+      ckeditor.config.overlayed.json: '{removePlugins: ''stylescombo,sourcearea,codemirror,youtube,table,tableresize,tableselection,tabletools,specialchar,image,image2,textselection,toolbar,link,indent,indentblock,indentlist,justify,hippopicker,elementspath''}'
```
wojtek-oledzki: @giacomolm - I think this will remove it from all HTML editors, right?

If true, can you let @ngta1 know how to disable it only for some content types?

repository-data/webfiles/src/main/resources/site/freemarker/publicationsystem/cilanding.ftl
```diff
@@ -1,15 +1,72 @@
...
+
 <section class="document-content">
     <#if document??>
         <h1 data-uipath="ps.document.title">${document.title}</h1>
         <div class="layout layout--large">
-            <div class="layout__item layout-2-3">
-                <@hst.html hippohtml=document.content />
+            <div class="layout__item layout-2-3" data-uipath="ps.cilanding.sections">
+                <#if document.subSections?has_content>
+                    <p>
```
wojtek-oledzki: we don't need that "<p>" here

repository-data/webfiles/src/main/resources/site/freemarker/publicationsystem/cilanding.ftl
```diff
@@ -1,15 +1,72 @@
...
+                                                    <a title="${attachment.text}" href="<@hst.link hippobean=attachment.resource/>">${attachment.text}</a> <span class="micro">&nbsp; [<@formatFileSize bytesCount=attachment.resource.length/>]</span>
+                                                </p>
+                                            </div>
+                                        </#list>
+                                    </div>
+                                </section>
+                            </#if>
+
+                            <#if section.assets?has_content>
+                                <p>
```
wojtek-oledzki: no need for "<p>" here. 

Is that what we want? No grey box?

repository-data/webfiles/src/main/resources/site/freemarker/publicationsystem/macro/resource-image.ftl
```diff
@@ -0,0 +1,36 @@
+<#ftl output_format="HTML">
+
+<#macro resourceImage file>
+    <#if file?ends_with(".xlsx") || file?ends_with(".xls")>
+        <svg width="24" xmlns="http://www.w3.org/2000/svg" class="media__img" viewBox="0 0 100 100">
```
wojtek-oledzki: please, use the same approach as in styleguide: `XLS`

repository-data/application/src/main/resources/hcm-config/namespaces/hippostd/html.yaml
ngta1: It can be global and field-specific.  Its currently global so will affect all RTE controls.  I can move that line to the individual field if required?  We use the RTE for other doc types like the About pages.

wojtek-oledzki: this change will affect "digital website" and "NIL" streams as well.

Might be worth checking with @giacomolm if you can define global "profiles" so you don't have to copy whole config to each content type.

repository-data/webfiles/src/main/resources/site/freemarker/publicationsystem/cilanding.ftl
ngta1: removed this p tag

repository-data/webfiles/src/main/resources/site/freemarker/publicationsystem/cilanding.ftl
ngta1: Yes, links will be separate from the grey box which will only contain attachments.

repository-data/application/src/main/resources/hcm-config/namespaces/hippostd/html.yaml
giacomolm: I would suggest locally to the specific document type, unless it's a global requirement. We did something similar [here|https://github.com/NHS-digital-website/hippo/blob/master/repository-data/application/src/main/resources/hcm-config/namespaces/website/service.yaml]

wojtek-oledzki: ace!

repository-data/webfiles/src/main/resources/site/freemarker/publicationsystem/cilanding.ftl
wojtek-oledzki: But "assets" and "attachments" are the same thing for end users, right?

Not sure if we should even support "attachments" in this content type.

ngta1: Assets can be (pre-uploaded) attachments, but also links to other CMS documents which the user selects on document completion.  I think as long as its ok for attachments uploaded as assets to be public before the document is actually published (which in this SHMI landing page case, it should be fine). I'll remove the Attachment compound type from this content type, and stick to assets instead.

ngta1:
Changes applied: removed attachment compound type, inline svg and CKeditor restriction applied to specific field using toolbarGroups.

repository-data/application/src/main/resources/hcm-config/namespaces/publicationsystem/cilandingasset.yaml
```diff
@@ -0,0 +1,117 @@
...
+            plugin.class: org.hippoecm.frontend.editor.plugins.field.NodeFieldPlugin
+            wicket.id: ${cluster.id}.field
+        jcr:primaryType: editor:templateset
+      /hipposysedit:nodetype:
+        /hipposysedit:nodetype:
+          /attachments:
+            hipposysedit:mandatory: false
+            hipposysedit:multiple: true
+            hipposysedit:ordered: false
+            hipposysedit:path: publicationsystem:Attachments-v3
```
wojtek-oledzki: no migration here, can we just call it `publicationsystem:Attachments` please

site/src/main/java/uk/nhs/digital/ps/beans/CiLandingAsset.java
```diff
@@ -0,0 +1,33 @@
...
+    public String getTitle() {
+        return getProperty("publicationsystem:title");
+    }
+
+    @HippoEssentialsGenerated(internalName = "publicationsystem:content")
+    public HippoHtml getContent() {
+        return getHippoHtml("publicationsystem:content");
+    }
+
+    @HippoEssentialsGenerated(internalName = "publicationsystem:Attachments-v3")
```
wojtek-oledzki: no need for `-v3`

ngta1:
Removed -v3 references. 1 attachment left in development data (...summary-hospital-level-mortality-indicator-shmi[1]...) as used in acceptance test, but removed others.

PR #186, opened 2018-02-08
[RPS-312] Added migration repository data module

jacsamell:
Added module for data created by the EXIM import to be inculded in
a distribution package that can be deployed to an environment to add
the migrated content.

Also fixed a load of hippo warnings.

### DIFF
--- a/ci-cd/Makefile
+++ b/ci-cd/Makefile
@@ -105,6 +105,11 @@ ondemand.upload:
 	echo "put ../target/website-$(VERSION)-on-demand-distribution.tar.gz /uploads/dists/$(VERSION).tar.gz" \
 		| sftp nhs@static.hosting.onehippo.com
 
+## Upload migration distribution file to ondemand sftp server
+ondemand.upload-migration:
+	echo "put ../target/website-$(VERSION)-distribution-with-migration-data.tar.gz /uploads/dists/$(VERSION)-with-migration-data.tar.gz" \
+		| sftp nhs@static.hosting.onehippo.com
+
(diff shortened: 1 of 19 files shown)


wojtek-oledzki:
hey @robValtech @giacomolm - this adds a functionality for creating a "migration distribution". Please take a look and comment.

jacsamell:
@giacomolm 
Yeah exactly, except we will bootstrap the content into an already deployed hippo instance by re-deploying with the migration data content as well.

giacomolm:
ok, thanks for clarification. If I have understood your process correctly, this is one of the approach we were going to suggest. The only question I have, is this process incremental ? What we had in mind was to export only part of the content, that was not part of previous migration. 

Example, if in a previous migration step you only imported a sub-folder **a**, _like publications/a_, and in the next phase you are going top import a subfolder **b**, what we had in mind was to export in that bootstrap module only the new content. Does it make sense for you?

jacsamell:
You wouldn't be able to do this incrementally currently as our process is to do a full migration with all the content and wipe all the migrated data before doing another full import. It may be possible to extend this approach to make it incremental but we've not looked into it.

giacomolm:
Ok,but now I don't understand why you created this module. By default the product automatically exports content to a bootstrap-repository module (unless the **content** auto-export is disabled). Coming back to my first comment, if you import content using exim and the auto-export is active, then all the new files will end up automatically into the bootstrap-repository module. 

I think you are aware of this, if I am missing something (sure I am) let me know

PR #187, opened 2018-02-08
[DW-0000] Changing name from article to services

giacomolm:
Hi guys, this pull request is changing the name of a sitemap item (and the relative content path) from article to services,


giacomolm:
@ziemsky thanks for checking. I refactored according to your comments. Just a curiosity, how the format-yaml is able to reformat based on the real changes?

PR #204, opened 2018-02-16
[RPS-339] Restricted folder access for editors and authors

jacsamell:
When we go live we only want the editors and authors to be able to
access the publications system folders. When the other domains go
live they will create new user groups for their users.

Also contains a few other minor fixes including:
- page not found page applied site wide
- added beta to the header
- updated application version


cms/src/main/java/uk/nhs/digital/ps/modules/FullTaxonomyModule.java
```diff
@@ -50,6 +51,12 @@ private void createFullTaxonomyProperty(Node document) {
             .map(StringValue::new)
             .toArray(Value[]::new);
 
+        // We have seen this happen when a document has only invalid taxonomy keys
```
wojtek-oledzki: :+1:

PR #206, opened 2018-02-19
[RPS-383] Order of datasets on a publication

jacsamell:
The datasets will be sorted with respect to the numbers prefixing
the title in logical order.

### DIFF
--- a/site/src/test/java/uk/nhs/digital/ps/components/DocumentTitleComparatorTest.java
+++ b/site/src/test/java/uk/nhs/digital/ps/components/DocumentTitleComparatorTest.java
@@ -1,0 +1,104 @@
+package uk.nhs.digital.ps.components;
+
+import static java.util.Arrays.asList;
+import static java.util.stream.Collectors.toList;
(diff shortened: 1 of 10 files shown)


site/src/test/java/uk/nhs/digital/ps/components/DocumentTitleComparatorTest.java
```diff
@@ -0,0 +1,100 @@
...
+                    "1.10.1 hippo",
+                    "2.0.0 antelope",
+                    "10 tiger",
+                    "10.0.9 elephant",
+                    "10.2.1 lion",
+                    "11.0.1 panther",
+                    "101 cheater",
+                    "hippo",
+                    "zebra"
+                )
```
ziemsky: Could perhaps use a case illustrating impact of mixing lower- and upper-case characters.

jacsamell: 👍

PR #210, opened 2018-02-20
[NIL-96] NIL landing page content

mchalejeiwc:
This change comprises of:

1) NIL "hub links" and landing pages doc types. The test instances of these are just skeletal.
2) New "action link" compound - this is to allow the configuration of multiple links within the NI hub. Similar to related link but without the protocol validation.
3) Changes to the NI hub template to support the above. The test content is a work in progress to match the evolving prototype - there will have to be a further iteration of this along with an acceptance test feature when that is signed off.

NB, I've added a few styles in to the main .less file. I'm not 100% sure where this should go.


mchalejeiwc:
Had to rebase/resolve conflicts, and then make a change to prevent UUIDs being regenerated, hence breaking the seeded data links. All done now.

PR #212, opened 2018-02-21
[RPS-89] Add kite mark for national statistics publications

ngta1:
Publications with an information type of national statistics
will have the national statistics kite mark logo displayed
next to it on the search results page and publications page.


acceptance-tests/src/test/resources/features/site/publication.feature
```diff
@@ -100,3 +100,16 @@ Feature: Ensure publication page displays required fields.
             | 1.9 Dataset  |
             | 1.10 Dataset |
             | 2.0 Dataset  |
+
+    Scenario: National Statistics kite mark on National statistic information type only
+        Given I navigate to the "search" page
+        When I search for ""
+        Then I click on the "National statistics" link
+        And I can click on the "Morbi tempor euismod vehicula" link
```
jacsamell: Do you check for the icon on the search page or just the publication page?
Probably worth checking both.

This would probably read better as Given -> When -> And -> Then ...

acceptance-tests/src/test/resources/features/site/publication.feature
```diff
@@ -100,3 +100,16 @@ Feature: Ensure publication page displays required fields.
...
+        And I should also see:
+            | Publication Information Types | National statistics           |
+        And I can see "National statistics" link
```
jacsamell: The kite mark is just an icon not a link right?

acceptance-tests/src/test/resources/features/site/publication.feature
ngta1: Will add a separate check on the search page too.

acceptance-tests/src/test/resources/features/site/publication.feature
ngta1: yes, its just an icon. I'll update the "I can see" method to also take "image" as well as link|button

PR #221, opened 2018-02-23
[RPS-392] Added switch for latest publication flag

jacsamell:
You will be able to turn off the latest version headers on the
series page and on the search results.

### DIFF
--- a/repository-data/application/src/main/resources/hcm-config/namespaces/publicationsystem/series.yaml
+++ b/repository-data/application/src/main/resources/hcm-config/namespaces/publicationsystem/series.yaml
@@ -55,9 +55,26 @@ definitions:
             wicket.css:
             - document-summary
             wicket.id: ${cluster.id}.left.item
+          /ShowLatest:
+            /cluster.options:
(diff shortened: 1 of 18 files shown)


repository-data/application/src/main/resources/hcm-config/namespaces/publicationsystem/series.yaml
```diff
@@ -55,9 +55,26 @@ definitions:
             wicket.css:
             - document-summary
             wicket.id: ${cluster.id}.left.item
+          /ShowLatest:
+            /cluster.options:
+              jcr:primaryType: frontend:pluginconfig
+            caption: Show First Publication as Latest
```
ngta1: Just wondering about wording for the end user.  Does "Show **First** Publication as Latest" mean "Show **most recent** publication as Latest"?  Should this field be ticked by default?

jacsamell: `Most recent` would be better 👍 
Left off by default as there are quite a few more that we want off than on.

PR #223, opened 2018-02-23
[RPS-391] Add Google analytics download and external link tracking

ngta1:
Added Google Analytics tracking events (using gtag) for download clicks
and external link clicks.  Events are displayed will in Google Analytics
dashboard under 'Behaviour > Events' or 'Real Time > Events' and grouped
into Category, Action and Label.

### DIFF
--- a/repository-data/webfiles/src/main/resources/site/freemarker/publicationsystem/dataset.ftl
+++ b/repository-data/webfiles/src/main/resources/site/freemarker/publicationsystem/dataset.ftl
@@ -100,13 +100,13 @@
         <ul data-uipath="ps.dataset.resources">
             <#list dataset.files as attachment>
                 <li class="attachment">
-                    <a title="${attachment.text}" href="<@hst.link hippobean=attachment.resource/>">${attachment.text}</a>;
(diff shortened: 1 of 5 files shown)


repository-data/webfiles/src/main/resources/site/freemarker/publicationsystem/dataset.ftl
```diff
@@ -100,13 +100,13 @@
         <ul data-uipath="ps.dataset.resources">
             <#list dataset.files as attachment>
                 <li class="attachment">
-                    <a title="${attachment.text}" href="<@hst.link hippobean=attachment.resource/>">${attachment.text}</a>;
+                    <a title="${attachment.text}" href="<@hst.link hippobean=attachment.resource/>" onClick="logGoogleAnalyticsEvent('Download attachment','<@fmt.message key='labels.dataset'/>','${attachment.resource.filename}');">${attachment.text}</a>;
```
jacsamell: Not sure we want `<@fmt.message key='labels.dataset'/>` to be the category key as this can be changed in the CMS which would mess up the categories in GA. Maybe would be better to hardcode these to static values?

ngta1: Yes, changing the value means a new "category" dimension is created in GA, and there will be a split of the data (i.e. "Data set" downloads and "Dataset" downloads.  In this case, they are still the same category so yes hardcoded value would be better. Nice spot.

PR #229, opened 2018-02-27
[RPS-379] Allow authors to copy documents

jacsamell:
Copied the hippo document workflow and updated so authors are
allowed to copy documents.

Also various fixes for unit tests.


repository-data/application/src/main/resources/hcm-config/configuration/modules/scxmlregistry.yaml
```diff
@@ -0,0 +1,7 @@
+---
+definitions:
+  config:
+    /hippo:configuration/hippo:modules/scxmlregistry/hippo:moduleconfig/hipposcxml:definitions/documentworkflow:
+      hipposcxml:source:
```
wojtek-oledzki: I will check tomorrow, we might need to add the ".meta.before" to ensure our workflow is before the default one.

README.md
```diff
@@ -93,6 +93,43 @@ This is to make the format consistent so when comparing changes to a previous ve
...
+
+This will go through all the modules and sub-modules and check for newer versions for any that are specified
+in the pom.
+
+### Potential issues when upgrading Hippo version
+
+This section should be kept up to date with all known changes we have made to the code base which require 
+consideration when changing parent Hippo version.
+
+#### Document Workflow
```
ziemsky: :+1:

README.md
```diff
@@ -187,12 +224,11 @@ tests in the system. In such a case, you can speed up improve your change-the-te
 running selected Cucumber scenario(s) rather than all.
 
 To do that:
-- tag selected scenario(s) with a [custom tag][Gherkin tags], e.g. `@RUN`, placing it in the line directly preceding
-  the scenario,
-- modify class `AcceptanceTest`, replacing value of attribute `tags` in the by `@CucumberOptions` annotation with
-  your custom tag.
+- tag selected scenario(s) with a [custom tag][Gherkin tags] `@WIP`, placing it in the line directly preceding
```
ziemsky: This probably means that the `~@WIP` tag should be removed from `AcceptanceTest` class, too.

The original idea behind `~@WIP` in `AcceptanceTest` was that `@WIP` could be used to flag scenarios that we (temporarily) _don't_ want to have executed (say, when we want to commit our in-flight changes but we don't want them to create noise by known failing tests breaking the build). The `@RUN` tag was meant to target scenarios that we want to run exclusively.

Since `@WIP` seems to have taken over the original `@RUN`'s role, the continuing presence of `~@WIP` in `AcceptanceTest` may cause confusion or even interfere with tests.

repository-data/application/src/main/resources/hcm-config/configuration/modules/documentworkflow.scxml
```diff
@@ -0,0 +1,596 @@
...
+
+    <!-- the initial no-document state is used to prevent entering the handle state if there is no document -->
+    <state id="no-document">
+        <!-- automatic transition to and continue in parallel handle state only when a document variant exists -->
+        <transition target="handle" cond="!workflowData.documents.isEmpty()"/>
+    </state>
+
+    <parallel id="handle">
+
+        <!--  -->
```
ziemsky: In the original this comment says `<!-- the atomic status state is used to report several statuses and info feedback to the invoking workflow -->` - has it been removed intentionally?

jacsamell: Good spot, no idea how I did that, have reinstated

README.md
jacsamell: Ah yes didn't know about the other WIP tag, have removed it

repository-data/application/src/main/resources/hcm-config/configuration/modules/scxmlregistry.yaml
jacsamell: Ah ok, not sure about that, this is what auto-export spat out. Let me know what you find 🙂

PR #234, opened 2018-03-05
[DW-105] Adding type field to general document type

giacomolm:


ziemsky:
No objections from me.

PR #236, opened 2018-03-05
[DW-23] Add List document type

robValtech:
Also includes UI implementation of first pass [DW-26]

### DIFF
--- a/cms/pom.xml
+++ b/cms/pom.xml
@@ -1,7 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd"
-         xmlns="http://maven.apache.org/POM/4.0.0"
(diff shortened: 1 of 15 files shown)


cms/pom.xml
```diff
@@ -152,12 +150,17 @@
             <scope>compile</scope>
         </dependency>
 
+        <dependency>
+      <groupId>org.onehippo.cms7</groupId>
+      <artifactId>hippo-plugin-content-blocks</artifactId>
+      <scope>compile</scope>
+    </dependency>
+
```
jacsamell: Alignment out of whack on this

cms/pom.xml
```diff
@@ -152,12 +150,17 @@
...
     </dependencies>
     <build>
         <finalName>cms</finalName>
         <resources>
             <resource>
-                <filtering>false</filtering>
```
jacsamell: What does removing this do?

giacomolm:
Thanks @jacsamell, I changed accordingly

PR #237, opened 2018-03-05
[RPS-394] HotFix for missing hidden fields

wojtek-oledzki:
This is a bit dirty, but:

* Content migrator package is the same as the "Taxonomy" daemon package.
* Private class "TaxonomyDaemonCopy" extends "FullTaxonomyModule" in
  order to do stay in sync with the current implementation.
* Minimal change to "AbstractDaemonModule" to allow injection of the
  JCR session object.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,8 @@ cache:
     - "$HOME/.m2"
 
 install:
-  - make commit-msg-check
+  - make check.commit-msg
+  - make check.yaml
 
 script:
   - make test.unit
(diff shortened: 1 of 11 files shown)


.travis.yml
```diff
@@ -22,7 +22,8 @@ cache:
     - "$HOME/.m2"
 
 install:
-  - make commit-msg-check
+  - make check.commit-msg
+  - make check.yaml
```
ziemsky: :+1:

PR #238, opened 2018-03-06
[NIL-90] NIL search facets and supporting changes

mchalejeiwc:
Addition of 3 new facets which relate to the NIL Indicator document type.

Most of the "noise" here is the moving of 4 searchable fields and taxonomy/classifiable out of the Topbar component and into the main Indicator document. This is required because Lucene requires a flat data-model.

Also added geographical location to Indicator and cleaned up the Indicator.ftl.

(Note, there is no migration script as there is no content to migrate at this point. Also the hippo migration project will be updated separately.)


wojtek-oledzki:
@mchalejeiwc rebase with master - this will fix chrome driver problems in site tests

PR #240, opened 2018-03-06
[RPS-417] Introduce S3Connector module

wojtek-oledzki:
We can now get instance of S3 connector module which gives us basic
method for object management.

```
S3Connector s3Connector = HippoServiceRegistry
    .getService(S3Connector.class);
```

### DIFF
--- a/cms/src/main/java/uk/nhs/digital/ps/modules/S3ConnectorServiceRegistrationModule.java
+++ b/cms/src/main/java/uk/nhs/digital/ps/modules/S3ConnectorServiceRegistrationModule.java
@@ -1,0 +1,64 @@
+package uk.nhs.digital.ps.modules;
+
+import com.amazonaws.auth.AWSCredentialsProvider;
+import com.amazonaws.auth.SystemPropertiesCredentialsProvider;
+import com.amazonaws.regions.Region;
+import com.amazonaws.regions.Regions;
+import com.amazonaws.services.s3.AmazonS3;
+import com.amazonaws.services.s3.AmazonS3Client;
+import org.onehippo.cms7.services.HippoServiceRegistry;
+import org.onehippo.repository.modules.AbstractReconfigurableDaemonModule;
+import org.onehippo.repository.modules.ProvidesService;
+
+import uk.nhs.digital.s3.S3Connector;
+import uk.nhs.digital.s3.S3ConnectorImpl;
+
+import javax.jcr.Node;
+import javax.jcr.RepositoryException;
+import javax.jcr.Session;
+
+@ProvidesService(types = S3Connector.class)
+public class S3ConnectorServiceRegistrationModule extends AbstractReconfigurableDaemonModule {
+
+    private final Object configurationLock = new Object();
+    private S3Connector s3Connector;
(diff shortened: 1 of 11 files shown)


wojtek-oledzki:
This is mainly for REVIEW for now. Please DO NOT MERGE just yet.

mchalejeiwc:
Looks great. How are the aws access key/secret actually picked up by the credentials provider?

cms/src/main/java/uk/nhs/digital/s3/S3ConnectorImpl.java
```diff
@@ -0,0 +1,85 @@
...
+
+    public boolean uploadFile(String objectPath) {
+        s3.putObject(new PutObjectRequest(bucketName, objectPath, createSampleFile()));
+
+        return true;
+    }
+
+    protected void connect() {
+        AWSCredentialsProvider provider = new SystemPropertiesCredentialsProvider();
+        s3 = new AmazonS3Client(provider);
```
ziemsky: It seems that it would be better to move this entire initialisation, together with the 'region' field (which is only needed for initialisation) outside of this class and into `uk.nhs.digital.modules.S3ConnectorServiceRegistrationModule#doInitialize`. An instance of `AmazonS3Client` (the `s3` reference could then be provided to this class through its constructor.

wojtek-oledzki:
@mchalejeiwc SystemPropertyProvider, take a look at `Makefile` line 52-53. This (aws.secretKey) will be set by OnDemand infra team

cms/src/main/java/uk/nhs/digital/ps/modules/S3ConnectorServiceRegistrationModule.java
```diff
@@ -0,0 +1,64 @@
...
+
+    private final Object configurationLock = new Object();
+    private S3Connector s3Connector;
+    private String s3Bucket;
+    private String s3Region;
+    private String s3Endpoint;
+
+    @Override
+    protected void doConfigure(final Node moduleConfig) throws RepositoryException {
+        synchronized (configurationLock) {
```
jacsamell: As discussed can we check why this is synchronized?

cms/src/main/java/uk/nhs/digital/ps/modules/S3ConnectorServiceRegistrationModule.java
```diff
@@ -0,0 +1,64 @@
...
+        Region reg = Region.getRegion(Regions.fromName(s3Region));
+        AWSCredentialsProvider provider = new SystemPropertiesCredentialsProvider();
+        AmazonS3 s3 = new AmazonS3Client(provider);
+        s3.setRegion(reg);
+
+        String endpoint = reg.getServiceEndpoint("s3");
+        if (!s3Endpoint.isEmpty()) {
+            endpoint = s3Endpoint;
+        }
+        s3.setEndpoint(endpoint);
```
jacsamell: ternary operator might be easier to read but I don't really mind 🙂 

`s3.setEndpoint(isEmpty(endpoint) ? s3Endpoint : endpoint);`

PR #241, opened 2018-03-06
[DW-147] Add Static UI v1.0

robValtech:
This adds static pages to the sitemap in development mode, so that UI
can be developed and styled independently from the main codebase.

Go to /static-ui to check the pages representing the 'ideal' look of the 
PSD designs


ziemsky:
The commit's comment says _'Once the styling is happy in this branch, it can be implemented in the
master code base'_ - does that mean this static content (`freemarker/common/static-ui`) is not intended to be merged into master and only live in a branch?

robValtech:
@ziemsky This was indeed a separate branch, but now the intention is that it lives on master. This is just a leftover from the original commit message.

PR #243, opened 2018-03-07
[RPS-424] Adds doc explaining hot-fix release

wojtek-oledzki:


wojtek-oledzki:
I will update the doc accordingly with a bit more explanation what a hot-fix is and a scenario to use this approach.

PR #245, opened 2018-03-08
[NIL-122] Display tagged taxonomy on Indicator summary

mchalejeiwc:
Simple display of taxonomy 'categories' on the NIL Indicator summary page - as per the prototype.


mchalejeiwc:
"are you sure you want to display all taxonomies though?" 

There is a possibility that this could change. At the moment though, the tags displayed should be consistent with what you would see, for example, if searching viewing on the search results page.

PR #248, opened 2018-03-08
[RPS-423] Fix for broken preview

jacsamell:
Preview doesn't work with the latest version of `com.fasterxml.jackson.core` so I've stuck to the version that works.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -225,23 +225,10 @@
                 <version>1.3.2</version>
                 <scope>test</scope>
             </dependency>
-
-            <dependency>
-                <groupId>com.fasterxml.jackson.core</groupId>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -225,23 +225,10 @@
...
-                <groupId>com.fasterxml.jackson.core</groupId>
-                <artifactId>jackson-annotations</artifactId>
-                <version>2.9.4</version>
-            </dependency>
-
             <dependency>
                 <groupId>com.amazonaws</groupId>
                 <artifactId>aws-java-sdk</artifactId>
-                <version>1.9.6</version>
+                <version>1.11.291</version>
```
ziemsky: :+1:

PR #249, opened 2018-03-12
[NIL-91] Move assurance date and include in sort

mchalejeiwc:
Move the assurance date field out of the "Topbar" component and into the main Indicator document. This is required for sorting which has to be top-level.

Also includes a couple of minor template fixes.

Note, I've kept the acceptance tests separate for NIL to keep them simple.

(I'm not migrating existing content as NIL content is still in a state of flux and will be wiped and re-imported. HOWEVER post approval please let me merge this).

### DIFF
--- a/site/src/main/java/uk/nhs/digital/common/components/SearchComponent.java
+++ b/site/src/main/java/uk/nhs/digital/common/components/SearchComponent.java
@@ -183,7 +183,8 @@ public class SearchComponent extends CommonComponent {
         String sortParam = getSortOption(request);
         switch (sortParam) {
             case SORT_DATE:
-                queryBuilder.orderByDescending("publicationsystem:NominalDate", "publicationsystem:Title");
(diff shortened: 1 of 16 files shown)


site/src/main/java/uk/nhs/digital/common/components/SearchComponent.java
```diff
@@ -183,7 +183,8 @@ private HstQuery buildQuery(HstRequest request) {
         String sortParam = getSortOption(request);
         switch (sortParam) {
             case SORT_DATE:
-                queryBuilder.orderByDescending("publicationsystem:NominalDate", "publicationsystem:Title");
+                queryBuilder.orderByDescending("publicationsystem:NominalDate","nationalindicatorlibrary:assuranceDate",
```
jacsamell: space after comma please

mchalejeiwc:
I've addressed the minor actions. I'll push the green button when the build completes.

PR #259, opened 2018-03-13
[DW-140] Implement breadcrumb

robValtech:
Also contains UI story [DW-141]


site/pom.xml
```diff
@@ -137,6 +137,12 @@
             <scope>compile</scope>
         </dependency>
 
+        <dependency>
+            <groupId>org.onehippo.forge</groupId>
+            <artifactId>breadcrumb</artifactId>
+            <version>${forge.breadcrumb.version}</version>
```
wojtek-oledzki: versions should be defined only in DependencyManager

robValtech:
@Wojtek we updated the codebase, so that the version is set in the DependencyManager as requested.

PR #264, opened 2018-03-14
[NIL-132] Adds feature tests for indicator summary page

hugosykes:


acceptance-tests/src/test/resources/features/site/indicator.feature
```diff
@@ -0,0 +1,44 @@
+
+Feature: National Indicator summary page
+
+    As a site user I should see a summary page of all the Indicator information
+    so I can view each and assess the usefulness of the indicator.
+
+    Scenario: Indicator Summary page - fields
+        Given I navigate to the "sample-indicator" page
+        Then I should see a page titled "People with stroke admitted to an acute stroke unit within 4 hours of arrival at hospital"
+        Then I can see element with id "assured-status" with content "Independently assured by IGB"
```
jacsamell: Is there a reason you've decided not to use the uipath instead of id to keep it consistent with the publications stuff?

hugosykes: I was seeking a solution that didn't involve putting lots of 'ps.document....' all over the nil stuff

jacsamell: You can use the same works as in the id if you want to but just stick it in `data-uipath` instead of `id`

acceptance-tests/src/test/java/uk/nhs/digital/ps/test/acceptance/steps/site/SiteSteps.java
```diff
@@ -79,6 +79,17 @@ public void iSeeLink(String linkTitle) throws Throwable {
         assertNotNull("Can see element", sitePage.findElementWithTitle(linkTitle));
     }
 
+    @Then("^I (?:can )?see element with data-uipath \"([^\"]+)\" with content \"([^\"]+)\"$")
+    public void iSeeElementWithUiPathAndContent(String uiPath, String content) throws Throwable {
+        assertNotNull("Can see element with data-uipath", sitePage.findElementWithUiPath(uiPath));
+        assertTrue(sitePage.findElementWithUiPath(uiPath).getText().contains(content));
```
jacsamell: Can you use assertContainsString instead? it gi you more debug logging if this assert fails

hugosykes: That method doesn't exist as far as I can see but I have changed it now

PR #272, opened 2018-03-16
[NIL-130] Update NIL landing page shape and content

mchalejeiwc:
Restructuring of the NIL landing page/hublinks/static content.

Basic gist is there is now a consolidated 'help' page, replacing three separate pages, and these contain resources (forms) for download.

Most of the noise here is the document type changes, and the initial seeded NIL static content.


repository-data/application/src/main/resources/hcm-content/content/documents/administration/national-indicator-library/headers.yaml
```diff
@@ -245,6 +257,9 @@
     - headers.interpretationGuidelines
     - headers.qualityStatement
     - headers.technicalSpecification
+    - headers.adviceForm
+    - headers.addForm
+    - headers.applyForm
```
jacsamell: Again, remember if you're adding to existing resource bundle this won't have any effect on non-local environments.

mchalejeiwc: yep, will add them manually

mchalejeiwc:
Attachment resources are intended as they are the initial application forms for download.

PR #273, opened 2018-03-19
NIL 129 tabbed search

mchalejeiwc:
This is ready for formal review now.

This PR provides tabbed search using 'scope' to determine which content will be surfaced. Scope derivation is now more elegant than the first iteration.

Tab structure/style follows the Gov site as has been agreed.

Note as there is no (as yet) 'services' or 'news' content, the acceptance test asserts this.

### DIFF
--- a/acceptance-tests/src/test/resources/features/site/search.feature
+++ b/acceptance-tests/src/test/resources/features/site/search.feature
@@ -145,6 +145,34 @@ Feature: Basic search
             | Series / Collection ( ... |
             | Methodology ( ...         |
 
+    Scenario: Clicking on the 'All' tab displays the results page with full, unfiltered result set
+        Given I navigate to the "search" page
+        When I can click on the "All" link
+        Then I should see the list with title "DOCUMENT TYPE" including:
+            | Publication ( ...         |
+            | Data set ( ...            |
+            | Series / Collection ( ... |
+            | Methodology ( ...         |
+
+    Scenario: Clicking on the 'Data and Information' tab displays the results page with publications, data sets, series and methodologies
+        Given I navigate to the "search" page
+        When I can click on the "Data and Information" link
+        Then I should see the list with title "DOCUMENT TYPE" including:
+            | Publication ( ...         |
+            | Data set ( ...            |
+            | Series / Collection ( ... |
+            | Methodology ( ...         |
+
+    Scenario: Clicking on the 'Systems and Services' tab displays no results because we don't yet include them in the search component
+        Given I navigate to the "search" page
+        When I can click on the "Systems and Services" link
+        Then I can see the search description matching "No results for filters"
+
+    Scenario: Clicking on the 'News and Events' tab displays no results because we don't yet include them in the search component
+        Given I navigate to the "search" page
+        When I can click on the "News and Events" link
+        Then I can see the search description matching "No results for filters"
+
     Scenario: Each document type label is correctly displayed in search results
         Given I navigate to the "home" page
         When I search for "Bare Minimum"
--- a/repository-data/webfiles/src/main/resources/site/freemarker/common/base-layout.ftl
+++ b/repository-data/webfiles/src/main/resources/site/freemarker/common/base-layout.ftl
@@ -1,7 +1,8 @@
 <#ftl output_format="HTML">
 <!DOCTYPE html>
 <#include "../include/imports.ftl">
-<@hst.setBundle basename="emails"/>
+<#include "macro/searchTabsComponent.ftl">
+<@hst.setBundle basename="emails, feature-toggles"/>
 
 <html lang="en">
   <head>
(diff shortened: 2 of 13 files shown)


repository-data/webfiles/src/main/resources/site/freemarker/common/searchtabs.ftl
```diff
@@ -0,0 +1,13 @@
+<#ftl output_format="HTML">
+<#include "../include/imports.ftl">
+<@hst.setBundle basename="month-names,facet-headers,facet-labels,searchtab-labels"/>
+
+<div data-uipath="ps.search-tabs">
+    <#assign tabLink = searchLink + query???then('?query=${query}&', '?') + 'sort=${sort}&' +'area=' >
+    
+    <ul class="tabs-nav" role="tablist">
+        <#list tabs as tab>    
+            <li class="${(area=='${tab}')?string("active", "")}"><a tabindex="0" href="${tabLink}${tab}" title="<@fmt.message key="searchtab.${tab}"/>"><@fmt.message key="searchtab.${tab}"/></a></li>
```
jacsamell: What does the tabindex=0 mean? Should this not increment for the tabs?

acceptance-tests/src/test/resources/features/site/search.feature
```diff
@@ -145,6 +145,25 @@ Feature: Basic search
             | Series / Collection ( ... |
             | Methodology ( ...         |
 
+    Scenario: Clicking on the 'Data and Information' tab displays the results page with full, unfiltered result set
```
ziemsky: As discussed, this suggests that there is no 'All' tab anymore.

site/src/main/java/uk/nhs/digital/common/components/SearchComponent.java
```diff
@@ -192,15 +204,46 @@ private HstQuery buildQuery(HstRequest request) {
...
+        final HippoBean base = RequestContextProvider.get().getSiteContentBaseBean();
+        final HippoBean newsFolder = base.getBean(FOLDER_NEWS);
+        final HippoBean publicationsFolder = base.getBean(FOLDER_PUBLICATIONS);
+        final HippoBean servicesFolder = base.getBean(FOLDER_SERVICES);
+        final HippoBean dataAndInformationFolder = base.getBean(FOLDER_DATAANDINFORMATION);
+        final HippoBean nationalIndicatorLibraryFolder = base.getBean(FOLDER_NIL);
+
+        List<HippoBean> scopeBeans = new ArrayList<HippoBean>();
+
+        if (areaParam.equals(SearchArea.NEWS)) {
```
jacsamell: If you're using ENUMs you might want to use a switch here instead, more explicit.

repository-data/webfiles/src/main/resources/site/freemarker/common/searchtabs.ftl
mchalejeiwc: Should be tab?index, will fix

site/src/main/java/uk/nhs/digital/common/components/SearchComponent.java
```diff
@@ -192,15 +204,46 @@ private HstQuery buildQuery(HstRequest request) {
...
+
+        return scopeBeans.toArray(new HippoBean[scopeBeans.size()]);
+    }
+
     private String getSortOption(HstRequest request) {
         return Optional.ofNullable(getAnyParameter(request, REQUEST_PARAM_SORT)).orElse(SORT_DEFAULT);
     }
 
+    private SearchArea getAreaOption(HstRequest request) {
+        return SearchArea.valueOf(Optional.ofNullable(getAnyParameter(request, REQUEST_PARAM_AREA)).orElse(AREA_DEFAULT).toUpperCase());
```
jacsamell: I'd use AREA_DEFAULT as the enum itself rather than converting it to a string and then back to an enum.

```
String area = getAnyParameter(request, REQUEST_PARAM_AREA)

return area == null ? AREA_DEFAULT : SearchArea.valueOf(area)
```

repository-data/webfiles/src/main/resources/site/freemarker/common/base-layout.ftl
```diff
@@ -88,11 +88,22 @@
           </div>
         </div><!--
         --><div class="layout__item layout-2-3">
+          <#if hstResponseChildContentNames?seq_contains("tabs") >
```
ziemsky: Small that this `if` block is, I think it warrants turning into a macro, which could then be applied here (lines 91-95) and in lines 102-106.

site/src/main/java/uk/nhs/digital/common/components/SearchComponent.java
ziemsky: Agreed. Granted, that's replacing one kind of evil with another but the switch/case will be a bit easier to read.

mchalejeiwc: I'm changing to use a switch ;)

site/src/main/java/uk/nhs/digital/common/components/SearchComponent.java
mchalejeiwc: Good call

repository-data/webfiles/src/main/resources/site/freemarker/common/base-layout.ftl
mchalejeiwc: Will do

mchalejeiwc:
Updated to use Switch over If/Else, extracted macro for the search bar, and enabled the 'feature toggle'. This defaults to 'true' for local and acceptance tests, so I'll need to pre-populate that to 'false' in PROD.

PR #274, opened 2018-03-19
[RPS-440] Move series to top of search results

jacsamell:
Added a new hidden field `common:searchRank` which will replace
relevance in search results to put series at the top. The rest of
the results (publications/datasets) will be ordered by date finally
with archives at the bottom.

Also cleaned up some unused code and added favicon to the base ftl


wojtek-oledzki:
Looks good, as discussed let's deploy to UAT first and let SMEs any anyone else interested decide if that's the desired search behaviour.

PR #278, opened 2018-03-19
[RPS-452] Simplify hst hosts configuration

wojtek-oledzki:
We now have single mount with no alias name defined, hence hosts cleanup
to bring cms and site host configuration to the same group config.


wojtek-oledzki:
not working as expected on UAT

PR #280, opened 2018-03-19
[DW-174] Link the header logo and breadcrumb 'Home' to 'root' URL

robValtech:


docs/what-if/deploy-to-uat.md
```diff
@@ -0,0 +1,38 @@
...
+6. Checkout the branch you want to deploy 
+
+		cd ci-cd
+
+		make clean build ondemand.upload ondemand.deploy ENV=tst RD_CONF=~/nhsd.rundec.env
+
+7. Now see the status here: 
+
+		https://deploys.onehippo.com/project/nhs/activity
+      		/hippo:moduleconfig:
```
wojtek-oledzki: from this line down, looks like copy and paste from another document :)

robValtech: I'm sorry @wojtek, I'm not sure if I understand what you're referring to. This is a 'how-to' document with the information we have gathered around deploying to UAT, so that we have a common understanding of the process.

What parts of the PR are you requesting to change please?

jacsamell: He means from here down in the document is talking about AutoExport, it looks like a copy paste error!

jacsamell: Also, should this read release to test rather than UAT? UAT release is automatic

robValtech: I updated the md file to reflect your requested changes. 'UAT' has been replaced with 'Test' in the file name and in context, and we updated the documentation, so hopefully it makes more sense now. Thanks for the feedback boys!

PR #291, opened 2018-03-21
[NIL-134] New groups, folders, permissions for NIL

mchalejeiwc:
Define domain, security groups and permissions for the National Indicator Library

### DIFF
--- a/repository-data/application/src/main/resources/hcm-config/configuration/queries/templates/new-national-indicator-library-folder.yaml
+++ b/repository-data/application/src/main/resources/hcm-config/configuration/queries/templates/new-national-indicator-library-folder.yaml
@@ -1,0 +1,30 @@
+---
+definitions:
+  config:
+    /hippo:configuration/hippo:queries/hippo:templates/new-national-indicator-library-folder:
+      /hippostd:templates:
+        /hippostd:folder:
+          hippostd:foldertype:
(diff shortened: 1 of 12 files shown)


repository-data/development/src/main/resources/hcm-config/security/national-indicator-library-author.yaml
```diff
@@ -0,0 +1,10 @@
+---
+definitions:
+  config:
+    /hippo:configuration/hippo:groups/national-indicator-library-author:
+      hipposys:description: Users of the national indicator library folders who are
+        authors
+      hipposys:members:
+      - ci-author
```
jacsamell: Worth adding a new user? ci-author = Clinical Indicator author

repository-data/application/src/main/resources/hcm-config/configuration/queries/templates/new-national-indicator-library-folder.yaml
```diff
@@ -0,0 +1,30 @@
+---
```
jacsamell: You need to apply this to the NIL folder right?

mchalejeiwc: I think I have this correct now - certainly it seems to work fine locally.

PR #292, opened 2018-03-21
[RPS-454] Added test for preview

jacsamell:
Added a test for the preview that checks if a document can be created
and previewed. In order for this to run against the production only
config I have created a new module called local that only contains
the host.yaml config that allows the site to be mounted on localhost.


acceptance-tests/src/test/resources/features/cms/preview.feature
```diff
@@ -0,0 +1,8 @@
+Feature: CMS Preview
+    @WIP
```
ziemsky: Is the annotation left in place intentionally?

jacsamell: Nope, thanks 😁

pom.xml
```diff
@@ -544,8 +544,8 @@
                                         <todir>${development-module-deploy-dir}</todir>
                                     </file>
                                     <file>
-                                        <file>${project.basedir}/repository-data/migration/target/website-repository-data-migration-${project.version}.jar</file>
-                                        <todir>${migration-module-deploy-dir}</todir>
+                                        <file>${project.basedir}/repository-data/local/target/website-repository-data-local-${project.version}.jar</file>
```
giacomolm: The local module is part of the cargo.run profile, this means will affect deploys on a local env. How the hosts development settings are merge with the one in the local module ?

wojtek-oledzki:
```
module:
  after: website-repository-data-local
  name: website-repository-data-development
```

jacsamell: There are no host configs in development now anyway so nothing to merge

jacsamell: Good shout on the readme 👍

jacsamell: That file doesn't exist on my branch:

jacsamell: Yeah, but it won't be in master when I merge! I've moved it to local module

giacomolm: Ok, I understand this time. So basically in a local environment there will be only one host, in this case localhost

jacsamell: Nice one, yeah 😄

giacomolm: sorry guys I really need an explanation about it.. If I am blocking something, then you can merge the PR (but I will not approve). Otherwise I need explanation on why you want to have an host configuration just containing the settings for localhost

jacsamell: No worries,

giacomolm: yeah I get that, but just using the profile without-development-data didn't work?

giacomolm: Ok, from my understanding the local hosts.yaml file defines only the configuration for localhost, while the development one has the configuration for all the environment. right?

What's the point of extracting the localhost?

wojtek-oledzki: @giacomolm please spend 5min checking branch RPS-454-preview-tests on your laptop and checking content of repository-data folders.

Once you do that, it will be clear that:

1. On Demand hosts are defined in "application" module (not in development)
1. development / test content is in "development" module
1. config needed for running hippo on localhost is in "local" module

Before this change (on master branch) you will find that:

1. On Demand hosts are defined in "application" module
1. development / test content and config for running on localhost is in "development" module

To catch a bug which we had few days back, and possible any other future bugs - aka Digital Website (common) channel broken and not working - we need to start the app locally with "application" module + minimal config to make it work on "localhost", hence the "local" module extraction.

PR #298, opened 2018-03-22
[RPS-460] Added user-admin user group

jacsamell:
This group will only be able to change user account settings plus
taxonomy and URL rewrite rules.


ziemsky:
I'm not worthy to object, so :+1:

PR #301, opened 2018-03-22
[DW-190] Adding ArticleValveTest and PublicationValveTest

silrodriguez:


wojtek-oledzki:
Ace! squash to one commit and it should be good to go @silrodriguez

PR #306, opened 2018-03-23
[DW-195] Cleanup hosts setup

wojtek-oledzki:
This ticket also fixes DW-167.


giacomolm:
Hi @wojtek-oledzki , I've created a specific branch for DW-167 (https://github.com/NHS-digital-website/hippo/tree/DW-167-add-production-host)

If you look at the configuration, I needed to define a different virtualhost group since in the cms is not possible to use the same hst:mountpoint for two or more virtualhost in the same group

wojtek-oledzki:
tested locally - site and cms works. Channel manager works as well. No errors in logs.

Added `hst:nochannelinfo: true` to "beta" and "www" to quiet down the logs.

Tested on:
* cms.digital.nhs.uk:8080/cms
* beta.digital.nhs.uk:8080
* www.digital.nhs.uk:8080
* digital.nhs.uk:8080

What errors did you see in the logs/application @giacomolm ?

giacomolm:
Probably the log you quiet down were telling something like "an hst:channel node is allowed to be used only by one single mount": IMO that's not a correct setup. I would suggest to use a different virtualhost group as I did for in the other branch. 

Moreover the cms location for the beta node is pointing to 'https://cms.digital.nhs.uk'. Not sure if the beta node is just an alias or a different machine, but I would make more clarity around it.

PR #310, opened 2018-03-26
[RPS-465] Add "unlocker" group

wojtek-oledzki:
Members of the "unlocker" group can unlock documents.

Also added 404 page for /login urls.


repository-data/local/src/main/resources/hcm-config/configuration/groups/editor.yaml
```diff
@@ -0,0 +1,6 @@
+---
+definitions:
+  config:
+    /hippo:configuration/hippo:groups/editor:
+      hipposys:members:
+      - unlocker
```
ziemsky: I'm guessing that this does _not_ mean that every member of group editor gets 'unlocker' privileges, right?

repository-data/application/src/main/resources/hcm-config/configuration/modules/documentworkflow.scxml
```diff
@@ -115,8 +115,8 @@
                         <!-- for a not-current editor the current editor (holder) is reported through the 'inUseBy' feedback -->
                         <hippo:feedback key="inUseBy" value="holder"/>
                     </if>
-                    <if cond="!editor and workflowContext.isGranted(draft,'hippo:admin')">
```
ngta1: Assume removing this doesn't affect admin unlock operation?

repository-data/local/src/main/resources/hcm-config/configuration/users/unlocker.yaml
```diff
@@ -0,0 +1,8 @@
+---
+definitions:
+  config:
+    /hippo:configuration/hippo:users/unlocker:
+      hipposys:active: true
+      hipposys:password: unlocker
```
ziemsky: Does this mean we get a default 'unlocker' user alongside the built-in 'author' and 'editor' in vanilla installations?

What's the impact on already established installations: does such a user get added there on upgrade as well, necessitating a manual intervention to disable it in production environment?

wojtek-oledzki: not vanilla - localhost only, for testing

repository-data/application/src/main/resources/hcm-config/configuration/modules/documentworkflow.scxml
wojtek-oledzki: it does :) hence admin in "unlocker" group by default

PR #323, opened 2018-03-29
[DW-185] Added ResolveLinks script

silrodriguez:


repository-data/application/src/main/resources/hcm-config/configuration/update/ResolveLinks.groovy
```diff
@@ -0,0 +1,98 @@
+/*
```
wojtek-oledzki: where is this groovy script used?

cms/src/main/java/uk/nhs/digital/ps/migration/ResolveLinksHelper.java
```diff
@@ -0,0 +1,39 @@
+package uk.nhs.digital.ps.migration;
```
wojtek-oledzki: this package will be deleted shortly after GOSS migration. If this helper should exist post 19th April, please move it to another package.

jacsamell:
Is this PR still needed?

jacsamell:
This has been lying around for ages, if someone needs this a new PR can be opened from the branch.

PR #329, opened 2018-04-03
[DW-216] Handle optional summary and section title in the General doc…

robValtech:


repository-data/webfiles/src/main/resources/site/freemarker/common/general-article.ftl
```diff
@@ -18,10 +18,15 @@
                     <hr>
                     <nav role="navigation">
                         <ol class="article-section-nav__list">
+                            <#if document.summary?? && document.summary?length gt 0>
```
jacsamell: FYI you can replace this with `document.summary?has_content`

PR #346, opened 2018-04-12
[DW-220] Implement new styling and layout in the search results page

robValtech:
Also contains [DW-235]


acceptance-tests/src/test/resources/features/site/facets.feature
```diff
@@ -7,6 +7,7 @@ Feature: Faceted search
     Scenario: All expected facets are displayed
         Given I navigate to the "home" page
         When I search for "zzFacets_Testzz"
+        And I click on the accept cookie banner
```
jacsamell: It might be worth adding the step to one of the generic calls like `I navigate to` or `I search for` if it is required for all tests. It's not really relevant for the scenario.

smahavithana: Hi Jacob. Good suggestion. We will add this as a separate fix. Thanks a lot!

jacsamell: 👍

PR #353, opened 2018-04-16
[RPS-434] Apply styling

dsayce:
We've had a go at applying the styles.
I've amended the local header sass to add in a table and sub headings.


dsayce:
Hey @robValtech 
Could you take a look at this one for us?
Thanks
Dave

PR #367, opened 2018-04-20
[RPS-434] Restyled the publication templates

jacsamell:


repository-data/webfiles/src/main/resources/site/freemarker/common/search-strip.ftl
```diff
@@ -12,7 +12,7 @@
             <input type="text" name="query" id="query" class="search-strip__input" placeholder="${buttonLabel}" value="${query!""}" aria-label="${buttonLabel}">
         </div>
         <div class="search-strip__table-cell search-strip__table-cell--button">
-            <button class="search-strip__button" aria-label="${buttonLabel}">
+            <button id="btnSearch" class="search-strip__button" aria-label="${buttonLabel}">
```
ziemsky: Assuming the `id` attribute is here so that the acceptance tests have something to target - could it be replaced with `data-uipath`?

jacsamell: Good shout

repository-data/webfiles/src/main/resources/site/freemarker/publicationsystem/publication.ftl
```diff
@@ -4,112 +4,149 @@
 <#include "./macro/pageIndex.ftl">
 <#include "./macro/publicationHeader.ftl">
 <#include "./macro/sections/imageSection.ftl">
+<#include "../common/macro/pubsBreadcrumb.ftl">
 <#assign formatFileSize="uk.nhs.digital.ps.directives.FileSizeFormatterDirective"?new() />
 <@hst.setBundle basename="publicationsystem.labels,publicationsystem.headers"/>
 <#-- @ftlvariable name="publication" type="uk.nhs.digital.ps.beans.Publication" -->
 
+<#-- Add meta tags -->
+<#include "../common/macro/metaTags.ftl">
```
robValtech: FYI - Just a heads up: At the moment the pubs templates don't populate the meta tags correctly, because the document object is not defined. Our implementation uses `document.title` and `document.seosummary` to populate the meta tags for SEO. There is no action for RPS for now, we'll implement a solution once we're done with the template transformation tasks.

PR #369, opened 2018-04-23
[RPS-0] Fixes for CMS tests

jacsamell:
Few fixes for CMS tests and other generic bug fixes.

### DIFF
--- a/repository-data/development/src/main/resources/hcm-config/main.yaml
+++ b/repository-data/development/src/main/resources/hcm-config/main.yaml
@@ -2,7 +2,6 @@
 definitions:
   config:
     /hippo:configuration/hippo:modules/autoexport/hippo:moduleconfig:
-      autoexport:enabled: false
       autoexport:modules:
         operation: add
         type: string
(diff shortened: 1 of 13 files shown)


site/src/main/java/uk/nhs/digital/ps/components/PublicationComponent.java
```diff
@@ -19,7 +19,7 @@
     private static final String KEY_FACTS_ID = "Key Facts";
     private static final String ADMIN_SOURCES_ID = "Administrative Sources";
     private static final String DATASETS_ID = "Data Sets";
-    private static final String RESOURCES_ID = "Recources";
```
ziemsky: 😄

repository-data/development/src/main/resources/hcm-config/main.yaml
```diff
@@ -2,7 +2,6 @@
 definitions:
   config:
     /hippo:configuration/hippo:modules/autoexport/hippo:moduleconfig:
-      autoexport:enabled: false
```
ziemsky: Just checking: is this disabled because we also have this clause in the `main.yaml` in module `local`?

jacsamell: Spot on

PR #371, opened 2018-04-23
[RPS-0] Fixing s3 resources failing to be made public

jacsamell:


jacsamell:
With a little help from Wojtek! I'll pass on the error to valtech for further investigation

PR #373, opened 2018-04-24
[RPS-473] Fix for automated scheduling of release

jacsamell:


jacsamell:
Yeah that's probably a good idea 😀

PR #386, opened 2018-04-27
[RPS-437] S3 Peformance test

ziemsky:
Cucumber-driven test has been added allowing to verify
impact of S3 transfers on CMS' and Site's performance.
Test can be run both against a local as well as remote
instances; usage described in `what-if/test-s3-performance.md`.
The test does not run automatically as part of the build,
it has to be triggered manually.

Base URLs for CMS and Site used in acceptance tests now
configurable via command line, arguments `-DcmsUrl` and
`-DsiteUrl` (some Make 'test' targets may need modifying
to make use of them but the one for S3 test already does).

Make serve.noexport now uses `run` goal, ensuring it uses
all the required command line args, including ones
specifying AWS credentials.

Version of Tomcat used in locally run instances has been
configured to match that in BloomReach onDemand environments.

WebDriverProvider.newWebDriver and WebDriversRepo now make
possible to simulate a number of concurrent users accessing
the application.

### DIFF
--- a/acceptance-tests/src/test/java/uk/nhs/digital/ps/test/acceptance/pages/LoginPage.java
+++ b/acceptance-tests/src/test/java/uk/nhs/digital/ps/test/acceptance/pages/LoginPage.java
@@ -1,20 +1,33 @@
 package uk.nhs.digital.ps.test.acceptance.pages;
 
+import static org.slf4j.LoggerFactory.getLogger;
+
 import org.openqa.selenium.By;
+import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.WebElement;
+import org.slf4j.Logger;
 import uk.nhs.digital.ps.test.acceptance.webdriver.WebDriverProvider;
 
 public class LoginPage extends AbstractCmsPage {
 
+    private static final Logger log = getLogger(LoginPage.class);
+
+    private static final By USERNAME_FIELD_SELECTOR = By.name("username");
+    private static final By PASSWORD_FIELD_SELECTOR = By.name("password");
+    private static final By SUBMIT_BUTTON_SELECTOR = By.name("p::submit");
+    private static final By LOGOUT_BUTTON_SELECTOR = By.className("hippo-logout");
+
     private PageHelper helper;
 
-    public LoginPage(final WebDriverProvider webDriverProvider, PageHelper helper) {
-        super(webDriverProvider);
+    public LoginPage(final WebDriverProvider webDriverProvider,
+                     final PageHelper helper,
+                     final String cmsUrl) {
+        super(webDriverProvider, cmsUrl);
         this.helper = helper;
     }
 
     public void open() {
-        getWebDriver().get(URL);
+        getWebDriver().get(getUrl());
 
         if (isLoggedIn()) {
             helper.findElement(By.className("hippo-user-icon")).click();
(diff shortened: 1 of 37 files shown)


acceptance-tests/src/test/java/uk/nhs/digital/ps/test/acceptance/pages/LoginPage.java
```diff
@@ -54,12 +121,19 @@ public String findLoginErrorMessage() {
         return helper.findChildElement(hippoLoginFeedbackPanel, By.tagName("span")).getText();
     }
 
+    private void waitForLoginToComplete(final WebDriver webDriver) {
+        findLogoutButton(webDriver);
```
jacsamell: This call isn't blocking, the logout button will be evaluated to null instantly as you are calling findOptionalElement

acceptance-tests/src/test/java/uk/nhs/digital/ps/test/acceptance/steps/TestDataSteps.java
```diff
@@ -31,6 +39,12 @@
     public void clearTestData() {
         log.debug("Disposing of test data.");
         testDataRepo.clear();
+
+        try {
+            FileUtils.deleteDirectory(acceptanceTestProperties.getDownloadDir().toFile());
```
jacsamell: Won't this delete the download directory on our PCs?

acceptance-tests/src/test/java/uk/nhs/digital/ps/test/acceptance/steps/cms/ps/CmsSteps.java
```diff
@@ -101,6 +124,7 @@ private void createPublishedPublication(final Publication publication) throws Th
     }
 
     // Scenario: New Publication screen ========================================================================
+
```
jacsamell: Why all the blank lines?

acceptance-tests/src/test/java/uk/nhs/digital/ps/test/acceptance/pages/LoginPage.java
ziemsky: Actually, in the context it's being used in, it will block and this has been verified to be happening. In  performance tests (the only context where this method is being used) a custom instance of WebDriver is being used, one with so called '[implicit wait](https://www.seleniumhq.org/docs/04_webdriver_advanced.jsp)' configured (as produced by `WebDriverProvider#newWebDriver`).

We don't use the implicit waits in 'normal' acceptance tests because we need more fine-grained control, but in performance tests their use allowed to reduce boilerplate code by delegating polling for conditions to the WebDriver itlsef.

jacsamell: Ah I see, my worry is that someone will use this method in normal acceptance tests thinking it will block and it won't under normal conditions. Is there any way to make this clearer do you think?

acceptance-tests/src/test/java/uk/nhs/digital/ps/test/acceptance/steps/TestDataSteps.java
ziemsky: I sure hope it will! :-D

This is a directory that, by default, is located under Maven 'target' directory and is where WebDriver downloads files to during download tests. As it's being located under 'target', the content gets deleted with each `mvn clean` but in performance tests I needed those files to be gone before each test run (and I'd run those tests frequently without doing `mvn clean`) so that files downloaded during current test wouldn't have name conflicts with those left around by previous runs.

acceptance-tests/src/test/java/uk/nhs/digital/ps/test/acceptance/steps/cms/ps/CmsSteps.java
ziemsky: Good spot. Lines left in by accident. I'm removing them now.

acceptance-tests/src/test/java/uk/nhs/digital/ps/test/acceptance/steps/TestDataSteps.java
jacsamell: Ah OK so it's different to our default download folder on our PC, got it 👍

acceptance-tests/src/test/java/uk/nhs/digital/ps/test/acceptance/pages/LoginPage.java
ziemsky: That is a good point and a valid concern though, on closer look, it seems that the non-blocking behaviour _is_ the one displayed by the original version as both the original version and my overload, rely on the same `PageHelper.findOptionalElement > PageHelper.findFirstOrNull > WebDriver.findElements` `PageHelper` methods.

It therefore appears that whichever version someone ends up using will continue to work as it used to, as my changes didn't really alter anything in the way `findLogoutButton` always worked.

Having said that, your observation prompted me to think that perhaps it would be better to change either the `waitForLoginToComplete` or the overloads of `findLogoutButton` to actually be blocking. Funny thing: git history shows that this method used to be blocking but was then changed to start using the non-blocking `PageHelper.findOptionalElement`. In the event that this was incidental, I'll give making it properly blocking and see what impact this has on the tests - both the acceptance and performance ones.

ziemsky: Eventually decided to just make the `waitForLoginToComplete` more explicit in its blocking (and make it blocking even should someone want to use it with non-waiting, default version of WebDriver).

The `findLogoutButton` method needs to remain non-blocking in normal acceptance tests due to its use in `isLoggedIn` method; switching the former to use `WebDriver.findElement` would make it throw exception on element not found, while `WebDriver.findElements` only returns `null` in such a case without throwing anything which is what we want here.

jacsamell: 👍  nice one

PR #390, opened 2018-04-30
[RPS-365] Create breadcrumbs for CI pages

ngta1:
Added breadcrumbs for CI content (series,publication,
datasets). If the document sits under one of the
clinical indicator folders (i.e. CCG, SHMI), then a
breadcrumb back to that CI landing page is also added
to the hierarchy.

Note: 2 separate breadcrumbs had to be created, one
which uses the old styling (breadcrumb-ci-old-styling)
and one which uses the new styling (breadcrumb-ci). Once
the dataset template has been updated with the new styling,
then breadcrumb-ci-old-styling can be removed, and replaced
with the same breadcrumb that the publication/series page
uses.

### DIFF
--- a/acceptance-tests/src/test/resources/features/site/ciBreadcrumb.feature
+++ b/acceptance-tests/src/test/resources/features/site/ciBreadcrumb.feature
@@ -1,0 +1,51 @@
+Feature: As a user navigating to clinical indicator content, I need logical breadcrumbs
+    to help me browse around the various (often related) content.
+
+    Scenario: Series, publication, dataset (not within ci folder) breadcrumb
+        Given I navigate to the "series with publication with datasets" page
+        Then I should see the list with title "Navigation" including:
+            | series-with-publication-with-datasets |
+        And I should see the list with title "Navigation" not including:
+            | Data and information                   |
+        When I navigate to the "series publication with datasets" page
+        Then I should see the list with title "Navigation" including:
+            | series-with-publication-with-datasets |
+            | series publication with datasets      |
+        And I should see the list with title "Navigation" not including:
+            | Data and information                   |
+        When I navigate to the "series-publication-with-datasets Dataset" page
+        Then I should see the list with title "Navigation" including:
+            | series-with-publication-with-datasets     |
+            | series publication with datasets          |
+            | series-publication-with-datasets Dataset  |
+        And I should see the list with title "Navigation" not including:
+            | Data and information                   |
+
+    Scenario: Dataset (within CI folder) breadcrumb
+        Given I navigate to the "dataset under CI folder" page
+        Then I should see the list with title "Navigation" including:
+            | Data and information           |
+            | CCG Outcomes - Indicator Set  |
+            | CCG series                    |
+            | CCG publication               |
+            | CCG dataset                   |
+
+    Scenario: CI landing page (i.e. SHMI landing) breadcrumb
+        Given I navigate to the "SHMI landing" page
+        Then I should see the list with title "Navigation" including:
+            | Data and information                               |
+            | Summary Hospital-level Mortality Indicator (SHMI) |
+
+    Scenario: NIL document (not within ci folder) breadcrumb
+        Given I navigate to the "sample-indicator" page
+        Then I should see the list with title "Navigation" including:
+            | People with stroke admitted to an acute stroke unit within 4 hours of arrival at hospital     |
+        And I should see the list with title "Navigation" not including:
+            | Data and information   |
+
+    Scenario: Legacy Publication (not within ci folder) breadcrumb
+        Given I navigate to the "legacy publication" page
+        Then I should see the list with title "Navigation" including:
+            | Legacy Publication    |
+        And I should see the list with title "Navigation" not including:
+            | Data and information   |
--- a/repository-data/development/src/main/resources/hcm-content/content/documents/corporate-website/publication-system/clinical-indicators.yaml
+++ b/repository-data/development/src/main/resources/hcm-content/content/documents/corporate-website/publication-system/clinical-indicators.yaml
@@ -1,0 +1,293 @@
+---
+/content/documents/corporate-website/publication-system/clinical-indicators:
+  /ccg-outcomes-indicator-set:
+    /archive:
+      hippo:name: Archive
+      hippostd:foldertype:
+      - new-statistical-publications-and-clinical-indicators-folder
+      - new-statistical-publications-and-clinical-indicators-document
+      hippotranslation:id: a1e0168e-2532-4116-a443-f146502b4dbf
+      hippotranslation:locale: en
+      jcr:mixinTypes:
+      - hippo:named
+      - hippotranslation:translated
+      - mix:versionable
+      jcr:primaryType: hippostd:folder
+    /current:
+      /1.1-national-ccg-outcomes:
+        /ccg-dataset:
+          /ccg-dataset[1]:
+            common:FacetType: dataset
+            common:searchRank: 3
+            common:searchable: true
+            hippo:availability: []
+            hippostd:state: draft
+            hippostdpubwf:createdBy: admin
+            hippostdpubwf:creationDate: 2018-04-20T19:44:52.537Z
+            hippostdpubwf:lastModificationDate: 2018-04-25T15:09:34.624Z
+            hippostdpubwf:lastModifiedBy: admin
+            hippotranslation:id: b5384c78-18af-42e1-95b1-1a6f5c5a6cd3
+            hippotranslation:locale: en
+            jcr:mixinTypes:
+            - hippotaxonomy:classifiable
+            - mix:referenceable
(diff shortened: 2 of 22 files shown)


site/src/main/java/uk/nhs/digital/ps/beans/navigation/CiBreadcrumbProvider.java
```diff
@@ -0,0 +1,122 @@
...
+ * series, datasets) as they have a very specific setup in the CMS.
+ * </p>
+ */
+public class CiBreadcrumbProvider  {
+
+    private static final Logger log = LoggerFactory.getLogger(CiBreadcrumbProvider.class);
+    private static final String PATH_CI_LANDING_PAGES = "publication-system/ci-hub/";
+    private static final String SEPARATOR = "/";
+
+    private static final Map<String, String> ciFolderToCiLandingMap = new HashMap<String, String>() {
```
jacsamell: I'm not sure I like how hard coded this is. Could we make it so the behaviour is that the CI Folder is the same as the CI Landing page? We map have to rename a few of the landing pages but this shouldn't be two much of a problem if we add a couple of redirects. Then we wouldn't have to change this code every time a landing page changes. See what @ziemsky thinks

site/src/main/java/uk/nhs/digital/ps/beans/navigation/CiBreadcrumbProvider.java
```diff
@@ -0,0 +1,122 @@
...
+                    ciBreadcrumbItems.add(createBreadcrumbItem(ctx, ciLandingPage));
+                    break;
+                }
+                folder = (HippoFolder) folder.getParentBean();
+            }
+
+            // If Publication or dataset, find parent links for breadcrumb
+            // (note: not required for series since series
+            // documents are already top level breadcrumb)
+            if (content.getClass().equals(Publication.class)) {
```
jacsamell: Can you change to `content instanceof Publication`

site/src/main/java/uk/nhs/digital/ps/beans/navigation/CiBreadcrumbProvider.java
```diff
@@ -0,0 +1,122 @@
...
+
+                // Is publication part of series?
+                HippoBean publicationParent =  ((Publication) content).getParentDocument();
+                if (publicationParent != null) {
+
+                    // Create Series navigation
+                    Series series = (Series) publicationParent;
+                    ciBreadcrumbItems.add(createBreadcrumbItem(ctx, series));
+                }
+            } else if (content.getClass().equals(Dataset.class)) {
```
jacsamell: Same as above

site/src/main/java/uk/nhs/digital/ps/beans/navigation/CiBreadcrumbProvider.java
ziemsky: I agree with @jacsamell that it would be good to not to have this hardcoded.

Following our discussion, let's:
* add an optional field to the `cilanding` doctype, one that would point to the corresponding folders; if Hippo allows liking docs to folders (unlikely) let's use that, if not, let's use a simple text field,
* instead of populating the map with hardcoded values, have a query looking for all documents of type `cilanding` page that have the 'linking field' populated and use those to populate the map,
* after release, establish the cilanding-folder links in production.

Walt and Pete need to be made aware of the function of the linking field (whatever form it takes) and the fact that it'll need populating right after the release.

site/src/main/java/uk/nhs/digital/ps/beans/navigation/CiBreadcrumbProvider.java
```diff
@@ -0,0 +1,122 @@
...
+
+        final HstRequestContext ctx = request.getRequestContext();
+        HippoBean content = ctx.getContentBean();
+        boolean isClinicalIndicatorDocument = false;
+        List<BreadcrumbItem> ciBreadcrumbItems = new ArrayList<>();
+
+        // Is this a CiLanding page?
+        if (ciFolderToCiLandingMap.containsValue(content.getName())) {
+            isClinicalIndicatorDocument = true;
+        } else {
```
ziemsky: A lot of logic happens in this `else` block, I wonder if it could be made a bit more easy to follow if a few blocks were moved to private methods:
* the entire `while` loop (together with the preceding `HippoFolder folder` line, moved to something like `addCiLandingPageBreadcrumbItem(List<BreadcrumbItem> ciBreadcrumbItems)`,
* content of `if (content.getClass().equals(Publication.class)) {` block moved to, say, `addPublicationBreadcrumbItem(List<BreadcrumbItem> ciBreadcrumbItems)`,
* content of `} else if (content.getClass().equals(Dataset.class)) {` block moved to `addDatasetBreadcrumbItem(List<BreadcrumbItem> ciBreadcrumbItems)`

...thus applying the [SLAP principle](https://hackernoon.com/object-oriented-tricks-6-slap-your-functions-a13d25a7d994).

acceptance-tests/src/test/resources/features/site/ciBreadcrumb.feature
```diff
@@ -0,0 +1,51 @@
+Feature: As a user navigating to clinical indicator content, I need logical breadcrumbs
+    to help me browse around the various (often related) content.
+
+    Scenario: Series, publication, dataset (not within ci folder) breadcrumb
+        Given I navigate to the "series with publication with datasets" page
+        Then I should see the list with title "Navigation" including:
+            | series-with-publication-with-datasets |
+        And I should see the list with title "Navigation" not including:
+            | Clinical Indicators                   |
+        Given I navigate to the "series publication with datasets" page
```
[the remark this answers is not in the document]
ngta1: Yep would read better - updated to use `When`

repository-data/development/src/main/resources/hcm-content/content/documents/corporate-website/publication-system/clinical-indicators.yaml
```diff
@@ -0,0 +1,293 @@
+---
+/content/documents/corporate-website/publication-system/clinical-indicators:
+  /ccg-outcomes-indicator-set:
```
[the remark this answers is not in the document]
ngta1: the Clinical Indicators part creates the folder which ccg-outcomes-indicator-set acceptance test content sits in. It aims to mimic one of the structures of the live CI folder setup (Clinical Indicators > CCG Outcomes Indicator Set > Archive/Current...etc)

site/src/main/java/uk/nhs/digital/ps/beans/navigation/CiBreadcrumbProvider.java
ngta1: As discussed, updated to make it more dynamic. New field in `ciLanding` added (let me know if better name suggestion for field 😬).

site/src/main/java/uk/nhs/digital/ps/beans/navigation/CiBreadcrumbProvider.java
ngta1: Updated

site/src/main/java/uk/nhs/digital/ps/beans/navigation/CiBreadcrumbProvider.java
ngta1: Good call.  Reads a lot clearer now 👍

site/src/main/java/uk/nhs/digital/ps/beans/navigation/CiBreadcrumbProvider.java
```diff
@@ -0,0 +1,156 @@
...
+
+    public CiBreadcrumb getBreadcrumb() {
+
+        loadCiLandingBeans();
+
+        List<BreadcrumbItem> ciBreadcrumbItems = new ArrayList<>();
+
+        if (ciLandingBeans.stream()
+            .anyMatch(i -> i.getName().equals(currentDocumentBean.getName()))) {
+            isClinicalIndicatorDocument = true;
```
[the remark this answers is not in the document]
ngta1: Agreed, updated to encapsulate in method

site/src/main/java/uk/nhs/digital/ps/beans/navigation/CiBreadcrumbProvider.java
```diff
@@ -0,0 +1,156 @@
...
+import java.util.*;
+
+/**
+ * <p>
+ * Default Hippo Breadcrumb plugin uses menu/folder structure to generate breadcrumbs,
+ * but this doesn't fit for clinical indicator content (ciLanding, publication,
+ * series, datasets) as they have a very specific setup in the CMS.
+ * </p>
+ */
+public class CiBreadcrumbProvider  {
```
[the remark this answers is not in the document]
ngta1: I've created a separate ticket to add this.

ngta1:
Query updated as discussed.  Thanks for input @jacsamell

PR #394, opened 2018-05-02
[DW-237] DW-238 GDPR transparency doctype and UI

robValtech:


repository-data/application/src/main/resources/hcm-content/content/documents/administration/website.yaml
```diff
@@ -1,6 +1,193 @@
...
+      hippostd:state: draft
+      hippostdpubwf:createdBy: admin
+      hippostdpubwf:creationDate: 2018-04-27T09:48:32.260Z
+      hippostdpubwf:lastModificationDate: 2018-04-27T12:47:52.496Z
+      hippostdpubwf:lastModifiedBy: admin
+      jcr:mixinTypes:
+      - mix:referenceable
+      jcr:primaryType: resourcebundle:resourcebundle
+      resourcebundle:descriptions:
+      - Label of 'Asset Reference Number'
```
ngta1: Assume this is correct as "Label of" instead of "Label for" which is what all the below use?

PR #395, opened 2018-05-02
[RPS-451] Script to revert conversion of s3 attachments

jacsamell:
This script will only be used if something goes wrong with the s3
migration.


repository-data/application/src/main/resources/hcm-config/configuration/update/Migration20180502RevertExternalAttachmentsToEmbedded.groovy
```diff
@@ -0,0 +1,47 @@
...
+
+import org.hippoecm.repository.util.JcrUtils
+import org.onehippo.repository.update.BaseNodeUpdateVisitor
+
+import javax.jcr.Node
+
+import static org.hippoecm.repository.HippoStdNodeType.NT_RELAXED
+
+/**
+ * Swap the pre uploaded attachment node to using the S3 resource instead of the embedded one
```
ziemsky: This comment probably needs to be the other way around (presumably copied from original script)?

PR #397, opened 2018-05-02
[RPS-451] Changed the s3 rollback script

jacsamell:
You can't add mixin types to already defined node definitions as you get
errors when deploying to onDemand. I have updated the rollback
script to add the required mixin if the error occurs instead
of in the node definition.

### DIFF
--- a/repository-data/application/src/main/resources/hcm-config/configuration/update/Migration20180502RevertExternalAttachmentsToEmbedded.groovy
+++ b/repository-data/application/src/main/resources/hcm-config/configuration/update/Migration20180502RevertExternalAttachmentsToEmbedded.groovy
@@ -1,9 +1,12 @@
 package org.hippoecm.frontend.plugins.cms.admin.updater
 
+import org.hippoecm.repository.HippoStdNodeType
 import org.hippoecm.repository.util.JcrUtils
 import org.onehippo.repository.update.BaseNodeUpdateVisitor
 
 import javax.jcr.Node
+import javax.jcr.nodetype.ConstraintViolationException
+
 /**
  * Revert the nodes from external attachments to embedded ones for if anything goes wrong with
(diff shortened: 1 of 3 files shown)


repository-data/application/src/main/resources/hcm-config/configuration/update/Migration20180502RevertExternalAttachmentsToEmbedded.groovy
```diff
@@ -29,7 +33,21 @@ class Migration20180502RevertExternalAttachmentsToEmbedded extends BaseNodeUpdat
...
+            // Because the old resource node type doesn't have the relaxed mixin type,
+            // we get an exception here because we have got properties on this node
+            // which are not allowed. To get around this, we suppress the exception and
+            // add the relaxed mixin afterwards manually
+            try {
+                resourceNode.setPrimaryType(NODE_TYPE_OLD_RESOURCE)
+            } catch (ConstraintViolationException cve) {
+                if (cve.getMessage().equals(CONSTRAINT_ERROR_MESSAGE)) {
+                    log.info("Suppressing Constraint error and adding relaxed mixin")
+                    resourceNode.addMixin(HippoStdNodeType.NT_RELAXED)
```
ziemsky: Given the exception was thrown when attempting to change the `primaryType` of the node and seeing that the type change is not re-attempted here - does the node type actually get changed despite the exception being thrown or is adding the `relaxed` mixin sufficient instead?

jacsamell: Surprisingly, the primary node type change still happens even though the exception is thrown! So we end up changing the primary type and setting the mixin both at the same time when the transaction gets committed.

PR #403, opened 2018-05-11
[RPS-475] Improve S3 connector config mechanism

ngta1:
S3 Connector config updated to make it clearer which settings are
set by system properties and which are set via JCR.

The following are now set as system properties only, since the
likelyhood of wanting to change these on the fly (via JCR) is very
unlikely, and potentially risky:
external.aws.bucket
external.aws.region
external.aws.s3.endpoint

The following are set via JCR. These values can be modified
on the fly in the console, but because they are part of
the bootstrapped config, any updated values are not persisted
opon a new release. To update the values permanently, update
the values in s3ConnectorRegistrationModule.yaml:
maxConcurrentDownloadsCount
downloadShutdownTimeoutInSecs
maxConcurrentUploadsCount
uploadShutdownTimeoutInSecs

### DIFF
--- a/cms/src/main/java/uk/nhs/digital/externalstorage/modules/S3ConnectorServiceRegistrationModule.java
+++ b/cms/src/main/java/uk/nhs/digital/externalstorage/modules/S3ConnectorServiceRegistrationModule.java
@@ -9,7 +9,6 @@ import com.amazonaws.client.builder.AwsClientBuilder;
 import com.amazonaws.regions.Regions;
 import com.amazonaws.services.s3.AmazonS3;
 import com.amazonaws.services.s3.AmazonS3ClientBuilder;
-import org.apache.commons.lang3.Validate;
 import org.onehippo.cms7.services.HippoServiceRegistry;
 import org.onehippo.repository.modules.AbstractReconfigurableDaemonModule;
 import org.onehippo.repository.modules.ProvidesService;
@@ -34,15 +33,7 @@ public class S3ConnectorServiceRegistrationModule extends AbstractReconfigurable
 
     private final Object configurationLock = new Object();
 
-    private String s3Bucket;
-    private String s3Region;
-    private String s3Endpoint;
-
-    private int downloadsMaxConcurrentCount;
-    private long downloadsShutdownTimeoutInSecs;
-
-    private int uploadsMaxConcurrentCount;
-    private long uploadsShutdownTimeoutInSecs;
+    private S3ConnectorServiceRegistrationModuleParams params;
 
     private ExecutorService downloadExecutorService;
     private ExecutorService uploadExecutorService;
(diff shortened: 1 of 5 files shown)


cms/src/main/java/uk/nhs/digital/externalstorage/modules/S3ConnectorServiceRegistrationModuleParams.java
```diff
@@ -1,74 +1,91 @@
...
+        // (hippo:configuration/hippo:modules/s3ConnectorRegistrationModule/hippo:moduleconfig)
+        downloadsMaxConcurrentCount = getLongJcrPropertyValue(moduleConfigNode, KEY_DOWNLOAD_MAX_CONC_COUNT);
+        downloadsShutdownTimeoutInSecs = getLongJcrPropertyValue(moduleConfigNode, KEY_DOWNLOAD_SHUTDOWN_TIMEOUT_IN_SECS);
+        uploadsMaxConcurrentCount = getLongJcrPropertyValue(moduleConfigNode, KEY_UPLOAD_MAX_CONC_COUNT);
+        uploadsShutdownTimeoutInSecs = getLongJcrPropertyValue(moduleConfigNode, KEY_UPLOAD_SHUTDOWN_TIMEOUT_IN_SECS);
+    }
+
+    private long getLongJcrPropertyValue(Node moduleConfigNode, String jcrPropertyKey) {
+        try {
+            return moduleConfigNode.getProperty(jcrPropertyKey).getLong();
```
ziemsky: Hmm, it seems that my suggestion to use `getLong()` here and to set the type of the timeout and count fields as `long` resulted in a lot of casting to `int` and back to `long` outside of this class. Sorry about that.

How about:
* changing the type of those fields and of values returned by the corresponding getter methods to `int`,
* eliminating all the casting outside of this class,
* if needed, only doing the casting once, within this mehtod, changing it to return `int` and renaming it to `getIntJcrPropertyValue`.

It seems that it should also be possible to avoid casting even within this method by using (needs testing to find out whether `getDecimal()` will work correctly for these properties):

```java
return moduleConfigNode.getProperty(jcrPropertyKey).getDecimal().intValue();
```

cms/src/main/java/uk/nhs/digital/externalstorage/modules/S3ConnectorServiceRegistrationModule.java
```diff
@@ -82,33 +73,25 @@ protected void doShutdown() {
...
+    private void readInConfiguration(final Node moduleConfigNode) throws RepositoryException {
+        params = new S3ConnectorServiceRegistrationModuleParams(moduleConfigNode);
     }
 
     private void validateConfiguration() {
 
         try {
-            Validate.notBlank(s3Bucket, "Required argument is missing: " + AWS_BUCKET.getJcrModulePropertyKey());
-            Validate.notBlank(s3Region, "Required argument is missing: " + AWS_REGION.getJcrModulePropertyKey());
+            Validate.notBlank(params.getS3Bucket(), "Required argument is missing: " + S3ConnectorServiceRegistrationModuleParams.KEY_AWS_BUCKET);
```
ziemsky: This and subsequent lines that refer to the constants came out quite long (my fault for not having come up with shorter class names 😉).

How about replacing them with one static import, so that the constants can then be referenced without having to qualify them with the class name?:
```java
import static uk.nhs.digital.externalstorage.modules.S3ConnectorServiceRegistrationModuleParams.*;
```

cms/src/main/java/uk/nhs/digital/externalstorage/modules/S3ConnectorServiceRegistrationModuleParams.java
```diff
@@ -1,74 +1,91 @@
...
+    public static final String KEY_AWS_REGION =  "externalstorage.aws.region";
+    public static final String KEY_AWS_S3_ENDPOINT =  "externalstorage.aws.s3.endpoint";
+
+    // JCR PROPERTIES KEYS
+    public static final String KEY_DOWNLOAD_MAX_CONC_COUNT =  "maxConcurrentDownloadsCount";
+    public static final String KEY_DOWNLOAD_SHUTDOWN_TIMEOUT_IN_SECS =  "downloadShutdownTimeoutInSecs";
+    public static final String KEY_UPLOAD_MAX_CONC_COUNT =  "maxConcurrentUploadsCount";
+    public static final String KEY_UPLOAD_SHUTDOWN_TIMEOUT_IN_SECS =  "uploadShutdownTimeoutInSecs";
+
+    S3ConnectorServiceRegistrationModuleParams(final Node moduleConfigNode) {
```
ziemsky: I know I initially suggested placing the logic extracting the values in the constructor, but thinking about it it would probably better to:
* move this logic into a static public factory method (`extractParameters`? `newInstance`?), taking the `Node` as its parameter),
* make the constructor take the full list of already extracted values,
* make the constructor private.

This leaves the constructor with no other logic than just populating the fields which is typically a good practice.

ngta1:
Good call @ziemsky. All updated.

cms/src/main/java/uk/nhs/digital/externalstorage/modules/S3ConnectorServiceRegistrationModuleParams.java
ngta1: Now updated.

PR #412, opened 2018-05-21
[RPS-525] Cleaned up Cihub doc type

jacsamell:
Also fixed travis CMS tests


Makefile
```diff
@@ -1,28 +1,12 @@
 include env.mk
+include vars.mk
```
ngta1: 👍

PR #416, opened 2018-05-22
[RPS-469] Added reverse date sorting for upcoming

jacsamell:
Upcoming publications will be return below any published ones
in search results and the date order will be reverse so you
see the next to be published first.

This is achived with a derived data field that is used for
the sorting.

### DIFF
--- a/cms/src/test/java/uk/nhs/digital/common/OrderedSearchDateDerivedDataFunctionTest.java
+++ b/cms/src/test/java/uk/nhs/digital/common/OrderedSearchDateDerivedDataFunctionTest.java
@@ -1,0 +1,155 @@
+package uk.nhs.digital.common;
+
+import static org.hamcrest.Matchers.contains;
+import static org.hamcrest.Matchers.hasEntry;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertThat;
+
+import org.apache.commons.lang3.tuple.MutablePair;
+import org.apache.jackrabbit.value.BooleanValue;
+import org.apache.jackrabbit.value.DateValue;
+import org.apache.jackrabbit.value.LongValue;
+import org.junit.Test;
+
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.util.*;
+import java.util.stream.Collectors;
+import javax.jcr.RepositoryException;
+import javax.jcr.Value;
+
+public class OrderedSearchDateDerivedDataFunctionTest {
+
+    private static final SimpleDateFormat DATE_FORMAT = new SimpleDateFormat("dd-MM-yyyy");
+    private static final String ORDERED_SEARCH_DATE = "orderedSearchDate";
+
+    private OrderedSearchDateDerivedDataFunction function = new OrderedSearchDateDerivedDataFunction();
+
+    @Test
+    public void upcomingDateConvertedToNegativeMillis() throws Exception {
+        // given
+        Map<String, Value[]> map = generateInputMap("20-03-2020", false);
+
+        // when
+        Map<String, Value[]> result = function.compute(map);
+
+        // then
+        long expectedValue = -parseDate("20-03-2020").getTimeInMillis();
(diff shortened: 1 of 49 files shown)


cms/src/test/java/uk/nhs/digital/common/OrderedSearchDateDerivedDataFunctionTest.java
```diff
@@ -0,0 +1,122 @@
...
+    private OrderedSearchDateDerivedDataFunction function = new OrderedSearchDateDerivedDataFunction();
+
+    @Test
+    public void testUpcoming() throws Exception {
+        Map<String, Value[]> result = testFunctionCompute("20-03-2020", false);
+
+        long expectedValue = -parseDate("20-03-2020").getTimeInMillis();
+
+        assertThat("Result is as expected", result, hasEntry(ORDERED_SEARCH_DATE, new Value[]{new LongValue(expectedValue)}));
+    }
```
ziemsky: This and other test methods could more immediately communicate what is the set up here and what is the actual action that is being tested.

How about changing the order of operations to follow the given/when/then sequence, thus making it more explicit that `testFunctionCompute` leads to the the actual action that we're testing here?

Adding comments `// given`, `// when`, `// then` would further clarify purpose of each block, especially that the `testFunctionCompute` method makes it less obvious that the actual call under test is the `function.compute(...)` one.

cms/src/test/java/uk/nhs/digital/common/OrderedSearchDateDerivedDataFunctionTest.java
```diff
@@ -0,0 +1,122 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.util.*;
+import java.util.stream.Collectors;
+import javax.jcr.Value;
+
+@RunWith(DataProviderRunner.class)
```
ziemsky: Looks like a left-over from an early cut as I don't think any test in this class actually uses the data provider?

cms/src/test/java/uk/nhs/digital/common/OrderedSearchDateDerivedDataFunctionTest.java
```diff
@@ -0,0 +1,122 @@
...
+    public void testNoDate() throws Exception {
+        assertNull(testFunctionCompute(null, true).get(ORDERED_SEARCH_DATE));
+
+        assertNull(testFunctionCompute(null, false).get(ORDERED_SEARCH_DATE));
+
+        assertNull(testFunctionCompute(null, null).get(ORDERED_SEARCH_DATE));
+    }
+
+    @Test
+    public void testOrder() throws Exception {
```
ziemsky: The block of `addData` gives an impression of setting up expectations, whilst, in fact, it performs the actual test by calling `OrderedSearchDateDerivedDataFunction.compute`.

Continuing the spirit of making the tests easier to read by more clearly represent the setup-test-assertion blocks (the given/when/then) and given that earlier test methods consistently make an explicit call to `testFunctionCompute`, how about:
* repurposing the map to only store inputs (date and 'publicly accessible' flag),
* making the first operation on the stream a `.map` call to `testFunctionCompute` to produce outputs.

cms/src/test/java/uk/nhs/digital/common/OrderedSearchDateDerivedDataFunctionTest.java
```diff
@@ -0,0 +1,122 @@
...
+
+        List<String> sortedDates = inputToOutput.entrySet().stream()
+            .sorted(Map.Entry.comparingByValue())
+            .map(Map.Entry::getKey)
+            .collect(Collectors.toList());
+
+        // We sort from highest to lowest so need to reverse
+        Collections.reverse(sortedDates);
+
+        // The order should be published first in descending order followed by upcoming in ascending date order
```
ziemsky: This comment could actually be turned into an assertion message, replacing generic 'Order is as expected' one, thus communicating the expectation both in the code and in the test failure output.

jacsamell:
Thanks @ziemsky. I have re-structured tests to make them more readable.

PR #420, opened 2018-05-27
[DW-0] Make side navigation rendering more consistent across templates

robValtech:
Create `sectionNav` macro so that all pages can use the same style
side navigation with links pointing to W3C valid section ids in any
document.

Create `slugify` function to generate valid slugs from not always
section valid section titles.

Create `getSectionNavLinks` function that returns valid links for the
section nav for valid document sections.

Implement the slugified conditinal section navigation rendering in the
following templates:
- General
- Service
- CiLanding
- Indicator
- Publication
- Legacy publication

Other improvements
- Add meta tag rendering to the `Indicator` template
- Remove duplicate aria-label from the `search-strip` template
- Remove duplicate <main> from the `homeland` template
- Add top padding to filter navigation menu
- Restore top border of `Further information` section in Service
template


repository-data/webfiles/src/main/resources/site/freemarker/common/macro/sectionNav.ftl
```diff
@@ -0,0 +1,17 @@
+<#ftl output_format="HTML">
+
+<#macro sectionNav links = [{ "url": "#section-anchor", "title": "Title"  }] title="Page contents">
```
jacsamell: I'd consider not giving these default values as you would never want to fall back on them in real life

repository-data/webfiles/src/main/resources/site/freemarker/publicationsystem/publication.ftl
```diff
@@ -50,49 +50,73 @@
...
+            <#if index?has_content && index?size gt 1>
+            <div class="column column--one-third page-block page-block--sidebar sticky sticky--top">
+                <@sectionNav getSectionNavLinks(index)></@sectionNav>
+            </div>
+            </#if>
 
             <div class="column column--two-thirds page-block page-block--main">
-                <div id="Summary" class="article-section article-section--summary no-border">
-                    <h2><@fmt.message key="headers.summary"/></h2>
+                <div id="${slugify(summaryHeader)}" class="article-section article-section--summary no-border">
```
jacsamell: The IDs for these need to be kept in sync with the IDs in the index list. e.g uk.nhs.digital.ps.components.PublicationComponent#SUMMARY_ID.

I suggest you update the statics in PublicationComponent to be valid slugified values and then hard code the IDs in this template to the same values rather than using the user modifiable resource bundle values

repository-data/webfiles/src/main/resources/site/freemarker/publicationsystem/publication.ftl
```diff
@@ -1,9 +1,9 @@
 <#ftl output_format="HTML">
```
jacsamell: If you are making changes to publication.ftl can you also update publicationPage.ftl as well please as the pageIndex rendering on both should be the same. (This was the reason for the pageIndex macro) Either update pageIndex macro or remove it and update both templates

repository-data/webfiles/src/main/resources/site/freemarker/common/macro/sectionNav.ftl
robValtech: Makes sense, I'll remove them

repository-data/webfiles/src/main/resources/site/freemarker/publicationsystem/publication.ftl
robValtech: Makes sense, I'll do that

repository-data/webfiles/src/main/resources/site/freemarker/publicationsystem/publication.ftl
robValtech: Cool

robValtech: @jacsamell I've updated both the `publication.ftl` and `publicationPage.ftl` files to implement the new sideNav and I got rid of the pageIndex partial, so all affected templates use the same mechanism to render the side navigation.

PR #422, opened 2018-05-31
[RPS-533] Implement funnel and scatter plot charts

ngta1:
Add support for additional HighChart formats:
Scatter plots
Funnel plots - essentially a scatter plot with control
limits (lower and upper) applied


ngta1:
Very nice @ziemsky 👍

PR #423, opened 2018-05-31
[DW-286] Upgrade product version to 12.3.0

giacomolm:


pom.xml
```diff
@@ -465,6 +469,8 @@
                     </execution>
                 </executions>
             </plugin>
+            <!-- Disable this plugin after 12.3.0 upgrade -->
```
jacsamell: Why?

giacomolm: I had some issues during the upgrade, specially while dealing with the custom jackrabbit version. I think you also had the same issues when this enforcer plugin has been introduced, but what you probably forced specific version of the dependencies directly in the project root pom file (this one). 

In my opinion the solution you adopted to make this maven plugin happy is not really a solution, because you may override dependency versions defined in the product pom(s). 

You need to consider that our product is tested and shipped with specific artifact versions and overriding those dependencies in the project is not really an option. I can understand if those dependencies are project related, but in case they are used by the product they shouldn't be overwritten. 

IMHO this enforcer plugin is making the project not easy to maintain/upgrade and there isn't really a cost/benefit.

ziemsky: Are you saying that the versions enforced by 'our' POM were actually different from those specified by the parent (Hippo) POM, @giacomolm? I'd think they should be the same, meaning that on upgrade, they should simply be upgraded in line with the parent POM. If they were different, this was probably not intentional.

jacsamell: What's your solution to fix the fact that some dependencies need higher versions of other dependencies? Without this plugin we could get method not found exceptions etc. when using the other dependency

grahamd99: @giacomolm hi - please can you comment here re possible alternative solutions for tackling dependency problems.  As I understand it the enforcer plug-in helps identifies these at compile-time, meaning they can be discussed and resolved, hence mitigating risk of them only be found at run-time.

grahamd99: @giacomolm @jacsamell As discussed, we've decided (I'm happy to own decision) to go ahead with use of enforcer for now with v12.3.0, i.e. keep manually updating dependencies identified by enforcer, in CMS pom.xml.  Thereafter, see what BR Engineering suggest for 12.4.0 onwards.

pom.xml
```diff
@@ -90,6 +91,8 @@
         <tomcatDistributionUrl>https://archive.apache.org/dist/tomcat/tomcat-8/v8.0.48/bin/apache-tomcat-8.0.48.tar.gz</tomcatDistributionUrl>
 
         <forge.folderctxmenus.version>3.0.1</forge.folderctxmenus.version>
+        <!-- TO BE REMOVED WHEN UPGRADING FROM 12.3.0 -->
+        <hippo.jackrabbit.version>2.16.1-h2</hippo.jackrabbit.version>
```
ngta1: In other words only applicable/used in 12.3.0?  If we then move to 12.4.0 or 13.0.0 for example, the line needs to be removed?  How about upgrade to 12.3.1?

jacsamell: Correct, this issue is fixed in 12.3.1 and above as reported by @giacomolm so any upgrade should remove it

ngta1: cool thanks. Also noticed, this has 2 commits?

jacsamell: Was half @giacomolm and half me so I left the two to show who did what 🙂

ngta1: 👍

PR #426, opened 2018-06-01
[RPS-531] Add attachments and links to series

jacsamell:
Also added taxonomy to the schema.org markup for publications and
datasets.

### DIFF
--- a/repository-data/webfiles/src/main/resources/site/freemarker/include/imports.ftl
+++ b/repository-data/webfiles/src/main/resources/site/freemarker/include/imports.ftl
@@ -8,6 +8,7 @@
 <#assign formatRestrictableDate="uk.nhs.digital.ps.directives.RestrictableDateFormatterDirective"?new() />
 <#assign formatCoverageDates="uk.nhs.digital.ps.directives.CoverageDatesFormatterDirective"?new() >
 <#assign formatDate="uk.nhs.digital.ps.directives.DateFormatterDirective"?new() >
+<#assign formatFileSize="uk.nhs.digital.ps.directives.FileSizeFormatterDirective"?new() />
 
 <@hst.defineObjects />
 
@@ -138,7 +139,7 @@
 
         <#local dateRangeData = dateRangeData + { "comparableStartDate": comparableStartDate, "comparableEndDate": comparableEndDate } />
     </#if>
-    
+
     <#return dateRangeData />
 </#function>
 
--- a/site/src/main/java/uk/nhs/digital/ps/beans/Dataset.java
+++ b/site/src/main/java/uk/nhs/digital/ps/beans/Dataset.java
@@ -89,10 +89,10 @@ public class Dataset extends BaseDocument {
         return HippoBeanHelper.getParentPublication(this);
     }
 
-    public List<List<String>> getTaxonomyList() {
+    public List<String> getFullTaxonomyList() {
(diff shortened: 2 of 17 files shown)


repository-data/webfiles/src/main/resources/site/freemarker/include/imports.ftl
```diff
@@ -8,6 +8,7 @@
 <#assign formatRestrictableDate="uk.nhs.digital.ps.directives.RestrictableDateFormatterDirective"?new() />
 <#assign formatCoverageDates="uk.nhs.digital.ps.directives.CoverageDatesFormatterDirective"?new() >
 <#assign formatDate="uk.nhs.digital.ps.directives.DateFormatterDirective"?new() >
+<#assign formatFileSize="uk.nhs.digital.ps.directives.FileSizeFormatterDirective"?new() />
```
ngta1: 👍

site/src/main/java/uk/nhs/digital/ps/beans/Dataset.java
```diff
@@ -89,10 +89,10 @@ public Publication getParentPublication() {
         return HippoBeanHelper.getParentPublication(this);
     }
 
-    public List<List<String>> getTaxonomyList() {
+    public String getFullTaxonomyList() {
         assertPropertyPermitted(PropertyKeys.TAXONOMY);
 
-        return HippoBeanHelper.getTaxonomyList(getKeys());
+        return String.join(",", HippoBeanHelper.getFullTaxonomyList(this));
```
ziemsky: While the Dataset bean is a model for view, emitting a comma-separated string from here feels like tying the model a bit too closely to just one form of presentation. Should the same list of taxonomy values be needed to be displayed using some more elaborate HTML structure at some point, we'd need either to add a new method here, laboriously break the comma-separated string into a list in the template or re-work this method to return a list once again.

Could the method continue to return a list but delegate formatting to Freemarker? I haven't tried it, but [join](https://freemarker.apache.org/docs/ref_builtins_sequence.html#ref_builtin_join) built-in looks like a promising candidate.

site/src/main/java/uk/nhs/digital/ps/beans/PublicationBase.java
```diff
@@ -115,6 +109,12 @@ public HippoBean getParentDocument() {
         return getPropertyIfPermitted(PropertyKeys.INFORMATION_TYPE);
     }
 
+    public String getFullTaxonomyList() {
+        assertPropertyPermitted(Dataset.PropertyKeys.TAXONOMY);
+
+        return String.join(",", HippoBeanHelper.getFullTaxonomyList(this));
```
ziemsky: See similar comment for `Dataset` regarding delegating the 'join' operation to Freemarker's [join](https://freemarker.apache.org/docs/ref_builtins_sequence.html#ref_builtin_join) built-in.

site/src/main/java/uk/nhs/digital/ps/beans/Dataset.java
jacsamell: Good suggestion, have updated

site/src/main/java/uk/nhs/digital/ps/beans/PublicationBase.java
jacsamell: Have updated

PR #427, opened 2018-06-01
[RPS-534] Validate charts data file

ziemsky:
Custom validator triggered on document save now verifies that the
provided input for Highcharts is valid by letting the parser try to
interpret the parameters (including the input file) and reporting
failure in a controlled manner if the parsing fails with exception.


cms/src/main/java/uk/nhs/digital/ps/chart/HighchartsConfigDerivedDataFunction.java
```diff
@@ -0,0 +1,135 @@
...
+    private static final String PARAM_KEY_CHART_CONFIG = "chartConfig";
+
+    private final HighchartsInputParser parser;
+    private final ObjectMapper objectMapper;
+
+    public HighchartsConfigDerivedDataFunction() {
+        this(HippoServiceRegistry.getService(HighchartsInputParser.class), createObjectMapper());
+    }
+
+    HighchartsConfigDerivedDataFunction(HighchartsInputParser parser, ObjectMapper objectMapper) {
```
jacsamell: We can remove this now?

cms/src/main/java/uk/nhs/digital/ps/chart/HighchartsConfigDerivedDataFunction.java
```diff
@@ -0,0 +1,135 @@
...
+    private static final String PARAM_KEY_TYPE = "type";
+    private static final String PARAM_KEY_TITLE = "title";
+    private static final String PARAM_KEY_Y_TITLE = "yTitle";
+    private static final String PARAM_KEY_CHART_CONFIG = "chartConfig";
+
+    private final HighchartsInputParser parser;
+    private final ObjectMapper objectMapper;
+
+    public HighchartsConfigDerivedDataFunction() {
+        this(HippoServiceRegistry.getService(HighchartsInputParser.class), createObjectMapper());
```
jacsamell: As discussed, how about we delegate the construction of dependencies to a factory class and then we can get rid of the Service

ziemsky: As subsequently discussed, the module and the parser service registration were left in place after all, seeing that they were correctly leveraging the purpose of `HippoServiceRegistry` which is to provide common dependencies to other classes (to validator and to workflow task in this case).

Having said that, the derived function has been replaced with a workflow task and, thanks to the fact that the task is now being created by the workflow action, it was possible to eliminate the constructor previously created just to allow injection of mocks during tests.

PR #436, opened 2018-06-05
[DW-255] Adding script copying general docs as news

giacomolm:
[DW-255] Discarding titles in case of concatenated sections


giacomolm:
Thanks @jacsamell !

PR #437, opened 2018-06-05
[RPS-500] Added equations to text sections

jacsamell:


repository-data/application/src/main/resources/hcm-config/namespaces/publicationsystem/textSection.yaml
```diff
@@ -30,8 +30,12 @@ definitions:
             wicket.id: ${cluster.id}.field
           /text:
             /cluster.options:
-              ckeditor.config.overlayed.json: '{ toolbarGroups: [ { name: ''basicstyles''
-                }, { name: ''list'' } ] }'
+              ckeditor.config.appended.json: '{ removePlugins: ''tableresize,tableselection,tabletools,hippopicker,youtube,table,specialchar'',
+                mathJaxLib: ''https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.4/MathJax.js?config=TeX-AMS_HTML''
```
ziemsky: Would it be easy to store this file within our codebase and serve it from there?

It feels that relying on externally hosted resource may be brittle whilst serving it together with the rest of 'our' static resources would ensure that it's always available when the rest of the site is.

repository-data/webfiles/src/main/resources/site/freemarker/publicationsystem/macro/sections/sections.ftl
```diff
@@ -3,6 +3,9 @@
 <#include "imageSection.ftl">
 <#include "related-linkSection.ftl">
 
+<!-- Set up equation support from mathjax -->
+<script type="text/javascript" async src="https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.4/latest.js?config=TeX-AMS_HTML"></script>
```
ziemsky: Same question as above, regarding storing the file within our codebase and serving it from our instances of Hippo.

repository-data/application/src/main/resources/hcm-config/namespaces/publicationsystem/textSection.yaml
jacsamell: It is possible although I'm not sure if it would be worth it. It would make things like upgrading the version more difficult. We also have a few other external dependencies that we fetch like this (Highcharts and JQuery) so arguably we should be hosting these ourselves as well but it might be more maintainable to fetch them externally? What do you think?

PR #440, opened 2018-06-08
[RPS-533] Implement funnel and scatter plot charts

ngta1:
Add support for additional HighChart formats:
Scatter plots
Funnel plots - essentially a scatter plot with control
limits (lower and upper) applied


cms/src/main/java/uk/nhs/digital/ps/chart/input/SpecialisedHighchartsInputParser.java
```diff
@@ -0,0 +1,8 @@
+package uk.nhs.digital.ps.chart.input;
+
+import uk.nhs.digital.ps.chart.ChartType;
+
+public interface SpecialisedHighchartsInputParser extends HighchartsInputParser {
```
jacsamell: What's the need for both the interface and the abstract class?

ziemsky: Good question.

This is to hide as many implementation details from the callers as possible, thus making it possible to change the parsers' implementation without having to change the callers.

One implementation detail the abstract class betrays to the caller is the fact that the specialised parsers deal with Excel spreadsheat (the 'Xlsx' bit in its name). The caller shouldn't care/be aware what exact format the input takes (could be CSV or anything else), all it cares about is passing the input in and passing the output out from the parser best equipped to handle specific chart type.

Even more important issue is making the 'delegating parser' aware that there is some hierarchy behind the specialised parsers. Should we, say, ever want to replace inheritance with composition (thus eliminating the abstract class) we'd have to update the caller as well. Admittedly, this wouldn't take _that_ much effort in this case, but these things do eventually add up.

Coding to interfaces is a general good practice for various reasons, especially when dealing with variations and I myself am often guilty of not putting enough effort into following it more often but this occasion felt like the right opportunity to ensure better isolation.

jacsamell: Sounds good, thanks 👍

PR #446, opened 2018-06-13
[DW-83] Adding daemon module creating url rewrite rules

giacomolm:


giacomolm:
Hi guys, thanks for your comments. Daemon module is a "repository managed component" that is basically implementing a cms service. In my case this class is also listening to HippoEventBus events: when editors publish documents containing the friendly urls compound, then the related URL rewrite rule must be created/updated. 

Your comments are mainly related to break things down and adding unit tests:
1) More specifically, which are those responsibilities (can you please make 3 examples)? How can I break things down?
2) About unit tests, could you make couple of examples related to the (sub)-functionality that you want  me to test?

Thanks a lot for attention

PR #454, opened 2018-06-14
[RPS-541] Implemented Highmaps

jacsamell:


cms/pom.xml
```diff
@@ -194,6 +194,12 @@
             <groupId>org.onehippo.forge.folderctxmenus</groupId>
             <artifactId>folderctxmenus-frontend</artifactId>
         </dependency>
+
+        <dependency>
+            <groupId>com.googlecode.json-simple</groupId>
+            <artifactId>json-simple</artifactId>
+            <version>1.1.1</version>
+        </dependency>
```
ziemsky: Following our discussion and comment against `GeoJsonMapParser` class - let's remove this dependency.

cms/src/main/java/uk/nhs/digital/ps/chart/input/HighmapsXlsxInputParser.java
```diff
@@ -0,0 +1,68 @@
...
+import uk.nhs.digital.ps.chart.AbstractHighchartsParameters;
+import uk.nhs.digital.ps.chart.HighmapsParameters;
+import uk.nhs.digital.ps.chart.model.HighmapsModel;
+import uk.nhs.digital.ps.chart.model.Series;
+
+import java.io.IOException;
+import java.util.Iterator;
+import javax.jcr.RepositoryException;
+
+public class HighmapsXlsxInputParser extends AbstractHighchartsXlsxInputParser {
```
ziemsky: It seems that this classes is missing unit tests.

PR #455, opened 2018-06-15
[RPS-518] Validate upcoming publication date

ziemsky:
From now on, Publication Date has to be no further than seven years in
the future and not in the past for Upcoming publications. There is no
such enforcement for Publication Dates of Finalised publications.

### DIFF
--- a/cms/src/main/java/uk/nhs/digital/ps/PublicationDateValidator.java
+++ b/cms/src/main/java/uk/nhs/digital/ps/PublicationDateValidator.java
@@ -1,0 +1,218 @@
+package uk.nhs.digital.ps;
+
+import static java.text.MessageFormat.format;
+import static java.time.temporal.ChronoUnit.DAYS;
+
+import org.apache.wicket.Session;
+import org.apache.wicket.model.IModel;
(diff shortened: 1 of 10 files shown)


cms/src/main/java/uk/nhs/digital/ps/PublicationDateValidator.java
```diff
@@ -0,0 +1,201 @@
...
+        // user outside of BST: 2018-01-14            gets saved in JCR as 2018-01-14T00:00:00Z
+        //
+        // In order to be able to easily compare the dates on day component (without taking time
+        // component into account as actual time component doesn't get persisted anyway),
+        // we need to compensate for potential daylight savings offset. Adding an hour to the
+        // value read from JCR before truncating the time component achieves that, bringing the
+        // 'decremented'/daylight saving dates into line while not affecting the non-daylight
+        // saving ones (actually, it does but the extra hour will get truncated anyway so it doesn't
+        // matter).
+        final LocalDate nominalDate = nominalDateRaw.toInstant()
```
jacsamell: Can you not just get the date after the time zone conversion which should then be correct?

PR #457, opened 2018-06-20
[DW-385] Update the contact details

robValtech:


robValtech:
@giacomolm - I hardcoded the content as per the original acceptance criteria. I know it's not ideal, but we'll have to deal with this later.

PR #460, opened 2018-06-21
[DW-256] Clean up `service` and `general` logic

robValtech:


robValtech:
Thank you @grahamd99. This is now on master!

PR #484, opened 2018-08-09
[DW-332] Add Published Work and Published Work Chapter document types

ngta1:
Includes DW-330 DW-229 DW-333 DW-484 DW-332 DW-323 DW-331 DW-322


grahamd99:
As per conversation, the scenario below causes URL re-write issue:
1) create parent published work doc type
2) add first resource (populated with description plus file), add second resource (no description, no file)
3) click save.  This is allowed (validation should spot the blank second resource, but it doesn't)
3) edit, add blank friendly URL
4) click save.  This is disallowed, as expected
5) remove blank second resource, click save and close.  This is allowed (validation should spot the blank friendly URL, but it doesn't)
6) click publish. URL re-write is created, with "Rewrite from" value (website then inaccessible)

PR #546, opened 2019-01-11
[DW-641] Projects functionality added

grahamd99:
Initially tried adding via Essentials.  That worked but it made a mess of formatting and order of lines in the various files.  I then re-tried using manual process here described here, which worked fine https://www.onehippo.org/library/enterprise/enterprise-features/workflow-process-management/add-workflow-process-management-to-a-project.html


grahamd99:
Also Lucene export added (https://www.onehippo.org/library/concepts/content-repository/add-the-lucene-export-service-to-a-project.html) to allow deployment of app to BRC

PR #548, opened 2019-01-14
[DW-670] Makes news and event have working compound filters

GeoffHayward:


GeoffHayward:
Rebased and includes a new change.

PR #565, opened 2019-02-01
[DW-706] Adds block link component and new icons

gazjoy:


gazjoy:
@GeoffHayward

The tests look to be failing in `publication.feature` The scenario being `Scenario: Display Resource Links, Attachments and Datasets in one list`

I split the new style links out into their own list so there are now 2 lists. I am thinking this is the likely cause.

PR #569, opened 2019-02-08
[DW-544] Adds sortable tables

gazjoy:


cms/src/main/resources/ckeditor/plugins/a11ychecker/myJs.js
```diff
@@ -0,0 +1,17 @@
+function () {
+    alert('my js!');
```
ngta1: Assume this file is no longer needed?

repository-data/local/src/main/resources/hcm-config/main.yaml
```diff
@@ -2,7 +2,6 @@
 definitions:
   config:
     /hippo:configuration/hippo:modules/autoexport/hippo:moduleconfig:
-      autoexport:enabled: false
```
ngta1: this file change/file can be excluded from the commit.

cms/src/main/resources/ckeditor/plugins/a11ychecker/myJs.js
gazjoy: Yep good catch!

repository-data/local/src/main/resources/hcm-config/main.yaml
gazjoy: Thanks @ngta1 will get this removed.

gazjoy: @ngta1 should this go in a `.gitignore`?

ngta1: We usually keep the file in as it sets autoexport to false by default.  Occasionally when we add a new document type/test data, we need autoexport to be switched on to spit out the YAML bootstrap files needed.

gazjoy:
@GeoffHayward I added to the source as the others do, I thought it was odd but this was consistent with the other js files. If there is any refactoring to do we should pick it up as a new story. It might be an idea to do some performance profiling on the frontend and measure these things. What do you think?

GeoffHayward:
Ok, excellent point.

PR #570, opened 2019-02-12
[DW-738] Adds a HTML and Head metadata injection point

GeoffHayward:


repository-data/application/src/main/resources/hcm-config/namespaces/website/service.yaml
```diff
@@ -226,6 +257,14 @@ definitions:
             hipposysedit:primary: false
             hipposysedit:type: String
             jcr:primaryType: hipposysedit:field
+          /string:
```
ngta1: Should this be called "metadata" instead of "string"?

PR #577, opened 2019-03-05
[DW-746] - Sortable table issues

esr360:


repository-data/webfiles/src/main/resources/site/freemarker/common/scripts/init-table-sort.js.ftl
```diff
@@ -1,12 +1,29 @@
 <#ftl output_format="HTML">
 
 <script>
-    var tableHead = document.querySelector('thead');
+    // @TODO - rather than targeting *all* tables, we should
```
ngta1: looks good, just wondering if this "TODO" still needs to be there or whether it can be removed?

esr360: The `TODO` has not been actioned, but in my opinion it's something that should be addressed at some point in the future.

PR #580, opened 2019-03-08
[DW-692] Replaces key facts infographics section in publication

esr360:


repository-data/webfiles/src/main/resources/site/freemarker/common/macro/fileInfographic.ftl
```diff
@@ -0,0 +1,10 @@
+<#ftl output_format="HTML">
+<#include "../../include/imports.ftl">
+<#include "component/infoGraphic.ftl">
+
+<#macro fileInfographic mimeType = "">
```
ngta1: Is this file used anywhere?

GeoffHayward: No idea. It could be a question for Gareth.

PR #604, opened 2019-04-26
[DW-762] Integrates highlighter.js with Hippo using Nashorn

GeoffHayward:


site/src/main/resources/META-INF/hst-assembly/overrides/valueListManager.xml
```diff
@@ -8,6 +8,7 @@
                 <entry key="rights" value="/content/documents/administration/website/rights"/>
                 <entry key="lawfulbasis" value="/content/documents/administration/website/lawful-basis"/>
                 <entry key="creditbanner" value="/content/documents/administration/website/news-banner-type"/>
+                <entry key="highlighterlanguages" value="/content/documents/administration/website/highlighter-languages"/>
```
ngta1: Wonder if this is used anymore?  If not, can remove.

GeoffHayward: You are correct, well spotted.

PR #678, opened 2019-07-19
[DW-768] Change UI site colours for accessibility

ngta1:


repository-data/webfiles/src/main/resources/site/scss/components/_expander.scss
```diff
@@ -63,6 +68,9 @@
         padding: 0 2px 0 0;
 
         details {
+
+
+
```
DamZiobro: You could consider to remove those 3 empty lines, but generally code looks good. I approve it.

ngta1: Thanks, now updated.

PR #682, opened 2019-07-23
[DW-897] Update styling of Service article and General article

DamZiobro:


repository-data/webfiles/src/main/resources/site/freemarker/common/general-article.ftl
```diff
@@ -4,6 +4,7 @@
 
 <#include "../include/imports.ftl">
 <#include "macro/sections/sections.ftl">
+<#include "macro/documentHeader.ftl">
```
ngta1: 👍

PR #712, opened 2019-08-29
[DW-314] Add blog post document type and UI

ngta1:
Includes a few updates for Cyber doc type value lists.

### DIFF
--- a/conf/query/lucene/indexing_configuration.xml
+++ b/conf/query/lucene/indexing_configuration.xml
@@ -179,6 +179,18 @@ https://wiki.apache.org/jackrabbit/IndexingConfiguration
         <property isRegexp="true" nodeScopeIndex="false">.*:.*</property>
     </index-rule>
 
+    <index-rule nodeType="website:blog">
+        <property>website:title</property>
+        <property>website:summary</property>
+        <property>website:shortsummary</property>
+        <property>website:seosummary</property>
+        <property>website:section</property>
+        <property>website:leadparagraph</property>
+        <property>website:backstory</property>
+
+        <property isRegexp="true" nodeScopeIndex="false">.*:.*</property>
+    </index-rule>
+
     <!-- hippo aggregates -->
     <aggregates>
         <nodetype>hippostd:html</nodetype>
--- a/repository-data/webfiles/src/main/resources/site/scss/components/_blog.scss
+++ b/repository-data/webfiles/src/main/resources/site/scss/components/_blog.scss
@@ -1,0 +1,130 @@
+.blog-author {
+    margin:0px;
+}
+
+.blog-backstory {
+    background-color: #E5E9EC;
+    padding: 20px;
+}
+
+.blog-contactus {
+    padding: 20px;
+    margin: 20px 0 40px ;
+    border-left: 5px solid #CDCBCE;
+}
+
+.blog-social {
+    display: flex;
+
+    .blog-social-icon {
+        padding: 5px 5px 5px 0px;
+        flex: none;
+
+        &__img {
+            width:80px;
+        }
+    }
+}
+
+.lead-paragraph p {
+    font-size: 22px;
+}
+
+.lead-image-container {
+    width: 100%;
+    margin-bottom: 30px;
+
+    .lead-image-caption {
+        padding: 5px 2px 5px 2px;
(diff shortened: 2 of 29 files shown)


repository-data/webfiles/src/main/resources/site/scss/components/_blog.scss
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    &__icon {
+        margin-right: 1.5rem;
+        width: 300px;
+        flex: none;
+
+        &__img {
+            width: 300px;
+            height: 200px;
```
DamZiobro: I fixing width and height good idea? I had similar problem in BlogHub and talked to Owain about it. He said we could fix width and crop on height. I implemented it that way using 'object-fit: cover' css. Is it worth to consider similar solution here?

repository-data/webfiles/src/main/resources/site/scss/components/_blog.scss
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        &__nhsimg {
+            width: 240px;
+            height: 140px;
```
DamZiobro: the same here

site/src/main/java/uk/nhs/digital/common/components/BlogComponent.java
```diff
@@ -0,0 +1,15 @@
...
+import org.hippoecm.hst.core.component.HstRequest;
+import org.hippoecm.hst.core.component.HstResponse;
+
+public class BlogComponent extends BaseGaContentComponent {
+
+    @Override
+    public void doBeforeRender(final HstRequest request, final HstResponse response) throws HstComponentException {
+        super.doBeforeRender(request, response);
+
+        request.setAttribute("currentUrl", request.getRequestURL().toString());
```
DamZiobro: Good to know we can do it this way ;)

repository-data/application/src/main/resources/hcm-config/configuration/update/EximImport.yaml
```diff
@@ -2,12 +2,13 @@
 definitions:
   config:
     /hippo:configuration/hippo:update/hippo:registry/EximImport:
-      hipposys:batchsize: 10
+      hipposys:batchsize: 1
       hipposys:description: Exim import script modified by Valtech
       hipposys:dryrun: false
       hipposys:parameters: "{\r\n  \"fileInJson\": true,\r\n  \"sourceBaseFolderPath\"\
-        : \"file:${java.io.tmpdir}/exim-import/\"\r\n}"
+        : \"file:/tmp/migrator/exim-import/\"\r\n}"
```
DamZiobro: Comment just to double check this change with fixed path is what we expect. Do we have the same fixed path for each test and production environment?

conf/query/lucene/indexing_configuration.xml
```diff
@@ -179,6 +179,18 @@ https://wiki.apache.org/jackrabbit/IndexingConfiguration
...
+        <property>website:summary</property>
+        <property>website:shortsummary</property>
+        <property>website:seosummary</property>
+        <property>website:section</property>
+        <property>website:leadparagraph</property>
+        <property>website:backstory</property>
+
+        <property isRegexp="true" nodeScopeIndex="false">.*:.*</property>
+    </index-rule>
+
```
DamZiobro: what does it actually do?

repository-data/webfiles/src/main/resources/site/scss/components/_blog.scss
ngta1: Good suggestion.  Will update.

repository-data/application/src/main/resources/hcm-config/configuration/update/EximImport.yaml
ngta1: This file not part of the PR, so will remove.

conf/query/lucene/indexing_configuration.xml
ngta1: The indexing-configuration allows to define which doc type/fields are to be included in the search.  The last line excludes everything else (so only index the fields defined).

DamZiobro: (y)

PR #750, opened 2019-10-15
[DW-432] Implement backend of Job Role doctype

DamZiobro:
+ [DW-1055] Add JobRole to person and blog doctype


site/src/main/java/uk/nhs/digital/website/beans/Role.java
```diff
@@ -17,21 +15,34 @@ public String getSectionType() {
...
+    }
+
     @HippoEssentialsGenerated(internalName = "website:primaryrole")
     public String getPrimaryrole() {
+        JobRole primaryRolePicker = (JobRole)this.getPrimaryrolepicker();
+        if (primaryRolePicker != null) {
+            return primaryRolePicker.getTitle();
+        }
         return getProperty("website:primaryrole");
     }
```
GeoffHayward: Does this method deviate from the auto generated bean? If it does I normally add `allowModifications = false` to the `@HippoEssentialsGenerated` annotation. I think doing so should stop your modification being lost.

site/src/main/java/uk/nhs/digital/website/beans/Role.java
```diff
@@ -17,21 +15,34 @@ public String getSectionType() {
...
 
+    @HippoEssentialsGenerated(internalName = "website:primaryroleorgpicker")
+    public HippoBean getPrimaryroleorgpicker() {
+        return getLinkedBean("website:primaryroleorgpicker", HippoBean.class);
+    }
+
     @HippoEssentialsGenerated(internalName = "website:primaryroleorg")
```
GeoffHayward: Does this method deviate from the auto generated bean? If it does I normally add allowModifications = false to the @HippoEssentialsGenerated annotation. I think doing so should stop your modification being lost.

site/src/main/java/uk/nhs/digital/website/beans/Role.java
DamZiobro: Yes - it looks like adding allowModifications=false seems to be reasonable here as we don't want BeanWriter to modify this function any way.

DamZiobro: I added allowModifications=false, tested and pushed updated changes.

PR #751, opened 2019-10-15
[DW-1017] Refactors Sticky Nav down to one method per type

GeoffHayward:
Types include:
- StickyNavSection (main one) used for jumping to sections
- StickyNavTags used for grouping by  tags
- StickyNavYers used for grouping by  years


site/src/main/java/uk/nhs/digital/ps/components/PublicationPageComponent.java
```diff
@@ -24,36 +16,7 @@ public void doBeforeRender(final HstRequest request, final HstResponse response)
         PublicationPage page = (PublicationPage) request.getRequestContext().getContentBean();
 
         request.setAttribute("page", page);
-        request.setAttribute("index", getIndex(page));
-        request.setAttribute("pageSections", pageSectionGrouper.groupSections(page.getBodySections()));
+        request.setAttribute("pageSections", pageSectionGrouper.groupSections(page.getSections()));
     }
 
-    private List<String> getIndex(PublicationPage page) {
```
DamZiobro: I like removing no longer needed code ;)

site/src/main/java/uk/nhs/digital/indices/StickySection.java
```diff
@@ -0,0 +1,28 @@
...
+     * @return A String that can be empty.
+     */
+    String getHeading();
+
+    /**
+     * The type of heading.
+     *
+     * @return The name of the heading level.
+     */
+    String getHeadingLevel();
```
DamZiobro: Could we also have default implementation of getHeading() and getHeadingLevel() - similar as implementation in Download.java? 

Thanks to that we will not need to dupliate the actual implementation in classes as Download, GallerySection etc. and similar future classes.

site/src/main/java/uk/nhs/digital/indices/StickySection.java
```diff
@@ -0,0 +1,28 @@
+package uk.nhs.digital.indices;
```
DamZiobro: Nice to have this prevention to

site/src/main/java/uk/nhs/digital/indices/IsStickySectionMainHeading.java
```diff
@@ -0,0 +1,25 @@
+package uk.nhs.digital.indices;
```
DamZiobro: Well done. Good to know we can implement Freemarker functions in Java.

repository-data/webfiles/src/main/resources/site/freemarker/common/blog-article.ftl
```diff
@@ -4,7 +4,7 @@
 
 <#include "../include/imports.ftl">
 <#include "macro/sections/sections.ftl">
-<#include "macro/sectionNav.ftl">
+<#include "macro/stickyNavSections.ftl">
```
DamZiobro: We can remove this macro from blog-article.ftl, as Blog article does not have sticky nav.

repository-data/webfiles/src/main/resources/site/freemarker/common/apimaster-article.ftl
```diff
@@ -81,6 +82,40 @@
         </div>
 
       </div>
-                          
+
 </article>
-                          
+
+<#-- Get section links in multiple elements -->
```
DamZiobro: I am not sure whether moving this function here is good idea. I know that this is only place where this function is used. However it was placed into imports.ftl as it is quite universal function and could be used in other places in the future.

If it was placed in imports.ftl it would be easier to find and reuse.

GeoffHayward: Is this comment pointing at the correct source? Only GitHub has the comment block on apimaster-article.ftl

site/src/main/java/uk/nhs/digital/indices/StickySection.java
GeoffHayward: I did think about this. But I deiced there was no clear default.

The best default would be `return getProperty("website:headinglevel");`. However this make too much of an assumption on the doc type.

Let me know what you think?

site/src/main/java/uk/nhs/digital/indices/IsStickySectionMainHeading.java
GeoffHayward: This was a real breakthrough in getting this ticket done. I will defiantly be a handy tool in the future too.

site/src/main/java/uk/nhs/digital/indices/StickySection.java
DamZiobro: We can leave it as it is for now. I see the 2 classes (Download, GallerySection) where actual default could work. If we will have more classes in the future we could refactor this part then.

repository-data/webfiles/src/main/resources/site/freemarker/common/apimaster-article.ftl
DamZiobro: It should be placed 1 line belo the comment. I meant about function 
<#function getNavLinksInMultiple sectionCompounds idprefix=''>

repository-data/webfiles/src/main/resources/site/freemarker/common/blog-article.ftl
GeoffHayward: Good spot

repository-data/webfiles/src/main/resources/site/freemarker/common/apimaster-article.ftl
GeoffHayward: I have moved that back now.

PR #754, opened 2019-10-23
[DW-408] Adds board doctype

GeoffHayward:


site/src/main/java/uk/nhs/digital/website/beans/Board.java
```diff
@@ -0,0 +1,54 @@
...
+@Node(jcrType = "website:board")
+public class Board extends CommonFieldsBean {
+
+    @HippoEssentialsGenerated(internalName = "website:quorate")
+    public Long getQuorate() {
+        return getProperty("website:quorate");
+    }
+
+    @HippoEssentialsGenerated(internalName = "website:shortsummary")
+    public String getShortsummary() {
```
DamZiobro: This property can be removed as it is already implemented in CommonFieldsBean.

PR #769, opened 2019-11-05
[DW-0] Fix NullPointerException in StickySection

DamZiobro:


DamZiobro:
This is the fix for this exception:

> 31.10.2019 12:00:55 +0000 [catalina-exec-11] WARN  [org.hippoecm.hst.core.container.HstComponentInvokerImpl.logComponentException():288] Component exception caught on window 'main' with component 'uk.nhs.digital.common.components.DocumentChildComponent': java.lang.NullPointerException at uk.nhs.digital.indices.StickySection.isMainHeading(StickySection.java:25)

PR #805, opened 2019-12-09
[DW-1157] Reverse the order of related news

GeoffHayward:


DamZiobro:
Change looks good, but it should be applied to PublicationBase.java, too

PR #826, opened 2020-01-30
[DW-1158] DW-1159 Implement `Update` doctype

robValtech:
- Implement in CMS with severity dropdown
- Implement in static UI (with styling)
- Implement as a reusable macro (with styling)
- Add to Publication, Legacy Publication, General and Service doc types

### DIFF
--- a/repository-data/webfiles/src/main/resources/site/freemarker/common/static-ui/shared/nav.ftl
+++ b/repository-data/webfiles/src/main/resources/site/freemarker/common/static-ui/shared/nav.ftl
@@ -3,34 +3,34 @@
 <nav class="static-ui-menu">
     <ul class="list list--reset">
         <li>
-            <a href="/static-ui/home">Home page</a>
+            <a href="<@hst.link path='/static-ui/home'/>">Home page</a>
         </li>
         <li>
-            <a href="/static-ui/service">Service document</a>
+            <a href="<@hst.link path='/static-ui/service'/>">Service document</a>
         </li>
         <li>
-            <a href="/static-ui/general">General document</a>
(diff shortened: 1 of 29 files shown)


repository-data/webfiles/src/main/resources/site/freemarker/common/static-ui/shared/nav.ftl
```diff
@@ -3,34 +3,34 @@
 <nav class="static-ui-menu">
     <ul class="list list--reset">
         <li>
-            <a href="/static-ui/home">Home page</a>
+            <a href="/site/static-ui/home">Home page</a>
```
GeoffHayward: These links will not work on the TST, DEV, etc. servers. You need to use somthing like:

```text
<a href="<@hst.link path='/static-ui/home'/>">Home page</a>
```

Then let the CMS resolve the path for the environment's config.

robValtech: This is done, and it works for the root `static ui` links. However when I do the same update to more nested routes such as `static-ui/component/block-link` in the `component/index.ftl` file, the links resolve as `/static-ui/component`.

GeoffHayward: Will they ever be put onto 'tst' or 'dev' for the P.O. to view it?

robValtech: We don't need it, this is only for local development. If there is a quick way of getting this working 'the right' way, let's do it, but I'll need your help with it, as my attempt didn't work :)

PR #863, opened 2020-03-05
[DW-1343] Fix PDF printing bug

robValtech:


repository-data/webfiles/src/main/resources/site/scss/nhsuk.scss
```diff
@@ -112,3 +112,10 @@
 @import "components/callout-box";
 @import "components/link-list";
 @import "components/utils";
+
+// Hide the PDF button if JS is off
+.no-js {
```
GeoffHayward: Is this the best file to put this? Wouldn't it be better placed in `nhsuk-print-pdf-document.scss` with something like the following: `.no-js & {
  display: none;
}`

robValtech: This is a fix for screen not printing, and it's the PDF printing button specifically we want to hide.

robValtech: I moved the definition into the `components/utils` file to keep the main sass file clean.

PR #897, opened 2020-04-02
[DW-1446] Create bloomreach search

mohammed-hasan-valtech:


site/components/src/main/java/uk/nhs/digital/intranet/components/SearchPageComponent.java
```diff
@@ -21,17 +47,90 @@ public SearchPageComponent(final GraphProvider graphProvider) {
...
+
+    protected int getCurrentPage(HstRequest request) {
+        return getAnyIntParameter(request, REQUEST_PARAM_PAGE, 1);
+    }
+
+    protected EssentialsListComponentInfo getComponentInfo(HstRequest request) {
+        return getComponentParametersInfo(request);
+    }
+
+    private List<SearchResult> getBloomreachResults(HstRequest request) {
```
alexandru-mitcu-valtech, line 80: Can you extract this into its own class and make it a bean?

alexandru-mitcu-valtech: I think it's going to be better if we separate them, and potentially use some `CompleteableFuture`s to run the 2 search queries at the same time

site/components/src/main/java/uk/nhs/digital/intranet/model/SearchResult.java
```diff
@@ -1,8 +1,10 @@
 package uk.nhs.digital.intranet.model;
 
-public interface SearchResult {
+import org.hippoecm.hst.content.beans.standard.HippoBean;
 
-    String getTitle();
+public interface SearchResult extends HippoBean {
```
alexandru-mitcu-valtech, line 5: Why does the interface need to extend `HippoBean` ? Is this required for the aggregation of results into `pageable` ?

alexandru-mitcu-valtech: Also, in this case, does `Person` need to extend `HippoItem` ?

mohammed-hasan-valtech: yes to point 1, for point 2, actually it could just implement the interface since we are overriding/implementing all the methods anyway

site/components/src/main/java/uk/nhs/digital/intranet/components/SearchPageComponent.java
mohammed-hasan-valtech: that makes sense

PR #899, opened 2020-04-02
[DW-1439] Add Full Width Image section

JodiWarren:
- Add Full Width Image compound document type
- Commit infographic changes
- Add empty FE templates for Full Width Image


repository-data/webfiles/src/main/resources/site/freemarker/common/macro/sections/fullWidthImage.ftl
```diff
@@ -0,0 +1,5 @@
+<#ftl output_format="HTML">
+
+<#macro fullWidthImageSection section>
+    <h1>Full Width Image goes here</h1>
```
GeoffHayward, line 4: Could this placeholder message end up being shown on production?

JodiWarren: @GeoffHayward Yes, but only if the section is invoked. It's a good point though. Luckily I have the Front End PR ready to go as soon as this is merged in, so we should be ok.

PR #909, opened 2020-04-08
[DW-1455] Create Statistics section back end

JodiWarren:
- Create Statistics compound type
- Create Statistics Item compound type
- Add number validation
- Add beans
- Add statistics to document types
- Add statistics to Sections macro

### DIFF
--- a/repository-data/application/src/main/resources/hcm-config/configuration/frontend/cms-validators.yaml
+++ b/repository-data/application/src/main/resources/hcm-config/configuration/frontend/cms-validators.yaml
@@ -54,6 +54,10 @@ definitions:
       jcr:primaryType: frontend:plugin
       plugin.class: org.hippoecm.frontend.editor.validator.plugins.RegExCmsValidator
       regex_pattern: ^(#[0-9A-Fa-f]{6})?$
+    /hippo:configuration/hippo:frontend/cms/cms-validators/statistics-valid-number:
+      jcr:primaryType: frontend:plugin
+      plugin.class: org.hippoecm.frontend.editor.validator.plugins.RegExCmsValidator
+      regex_pattern: (?:^\d+\/{1}\d+$)|(?:^\d?\.?\d+$)
     /hippo:configuration/hippo:frontend/cms/cms-validators/twitterusername:
       jcr:primaryType: frontend:plugin
       plugin.class: org.hippoecm.frontend.editor.validator.plugins.RegExCmsValidator
--- a/repository-data/application/src/main/resources/hcm-config/configuration/translations/cms.yaml
+++ b/repository-data/application/src/main/resources/hcm-config/configuration/translations/cms.yaml
@@ -34,6 +34,8 @@ definitions:
       publicationsystem-title: Title must be 250 characters or less
       rgbcolor: Must be RGB-based formatted color (#XXXXXX) where X is value from
         ranges [0-9] or [A-F]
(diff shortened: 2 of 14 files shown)


repository-data/application/src/main/resources/hcm-config/configuration/frontend/cms-validators.yaml
```diff
@@ -54,6 +54,10 @@ definitions:
       jcr:primaryType: frontend:plugin
       plugin.class: org.hippoecm.frontend.editor.validator.plugins.RegExCmsValidator
       regex_pattern: ^(#[0-9A-Fa-f]{6})?$
+    /hippo:configuration/hippo:frontend/cms/cms-validators/statistics-valid-number:
+      jcr:primaryType: frontend:plugin
+      plugin.class: org.hippoecm.frontend.editor.validator.plugins.RegExCmsValidator
+      regex_pattern: ^\d*(?:(?!,|\/\/).)(\d+\/{1}\d+|\d*)$
```
GeoffHayward, line 60: This regex matches 'T123' which should have been a non-match.

My test cases were:

 Allowed
----
1234
123.4
1234
123/4
.4
/123

Disallowed 
---
123//4
T123.4
T123
/123/
123,4

Try this regex: `^(?=^[^/]*/[^/]*$)[\d/.]+$`. This first checks that there is only one forward-slash and then goes on to validate that the characters are in the allowed set. Let me know if you need more help with integrating this regex.

repository-data/development/src/main/resources/hcm-config/main.yaml
```diff
@@ -2,7 +2,6 @@
 definitions:
   config:
     /hippo:configuration/hippo:modules/autoexport/hippo:moduleconfig:
-      autoexport:enabled: false
```
GeoffHayward, line 5: You probably should not committed the deletion of `autoexport:enabled: false`.

repository-data/application/src/main/resources/hcm-config/configuration/frontend/cms-validators.yaml
JodiWarren: @GeoffHayward Thanks, that's a good catch. I had updated this in the next PR, my bad on not bringing this forward!

That's almost spot on, I just want to update the allowed/disallowed list:

## Allowed
1234
123.4
1234
123/4
.4
~/123~

## Disallowed
123//4
T123.4
T123
/123/
123,4
**/123**

I've matched that with `(^\d+\/{1}\d+$)|(^\d?\.?\d+$)`: https://regexr.com/52hcp

GeoffHayward: Thanks for the link. I would make the capture groups non-capture groups, like this `(?:^\d+\/\d+$)|(?:^\d*\.?\d+$).`

GeoffHayward: Mind you, should `123.4/5` or `123.4/5.678` also match?

JodiWarren: Oh, good idea with the non-capture. I only put them into groups for legibility, so that helps clarify that.

I checked with CB and OD, and they confirmed we don't need decimal points combined with fractions

PR #910, opened 2020-04-09
[DW-1417] create project update and feed

mohammed-hasan-valtech:


repository-data/webfiles/src/main/resources/site/freemarker/common/project-update-feed.ftl
```diff
@@ -0,0 +1,188 @@
...
+<#--        </#if>-->
+<#--    </#list>-->
+
+    <#return links />
+</#function>
+
+<#function getTypeOfupdateLinks>
+    <#assign links = [] />
+
+<#--    <#list pageable.items?sort_by("typeOfUpdate") as key>-->
```
GeoffHayward, line 43: This bit looks strange. Is it commented out?

josephcyrus-valtech: it is commented out. It was waiting for some BE work but it has since been agreed to go ahead without, so I will remove this.

PR #912, opened 2020-04-14
[DW-1456] Add Statistics markup

JodiWarren:
- Add Statistics Section
- Add Statistics styles
- Add JS for number count-up


repository-data/application/src/main/resources/hcm-config/configuration/frontend/cms-validators.yaml
```diff
@@ -54,6 +54,10 @@ definitions:
       jcr:primaryType: frontend:plugin
       plugin.class: org.hippoecm.frontend.editor.validator.plugins.RegExCmsValidator
       regex_pattern: ^(#[0-9A-Fa-f]{6})?$
+    /hippo:configuration/hippo:frontend/cms/cms-validators/statistics-valid-number:
+      jcr:primaryType: frontend:plugin
+      plugin.class: org.hippoecm.frontend.editor.validator.plugins.RegExCmsValidator
+      regex_pattern: (^\d+\/{1}\d+$)|(^(\d|\.)+$)
```
GeoffHayward, line 60: This does not match '/123'. Also, this regex has three unused capture groups, where non-capture groups would be better placed.

repository-data/development/src/main/resources/hcm-config/main.yaml
```diff
@@ -2,7 +2,6 @@
 definitions:
   config:
     /hippo:configuration/hippo:modules/autoexport/hippo:moduleconfig:
-      autoexport:enabled: false
```
GeoffHayward, line 5: This deletion probably should not be committed.

PR #914, opened 2020-04-15
[DW-1387] Services and resources tile section build

ShaneAhern-IRE:
[DW-1387] Adding Cyber Security Services page + tile section builder


repository-data/development/src/main/resources/hcm-config/main.yaml
```diff
@@ -4,7 +4,8 @@ definitions:
     /hippo:configuration/hippo:modules/autoexport/hippo:moduleconfig:
       autoexport:enabled: false
       autoexport:modules:
-        operation: add
-        type: string
-        value:
-        - repository-data/development
+      - repository-data/development:/
```
ShaneAhern-IRE, line 7: Adding auto export modules to allow hst config to be exported correctly.
Currently, hst configuration is not exported locally when developing.

GeoffHayward: I worry about it changing the workflow.

PR #917, opened 2020-04-15
[DW-1379] Added announcement doctype

alexandru-mitcu-valtech:
[DW-1379] Added Java bean

[DW-1379] Converted expiryDate to datetime

[DW-1379] Added Announcements main folder

[DW-1379] Added new-intranet-document query
[DW-1379] Added BR config and freemarker template for Announcements

[DW-1379] Restricted allowed doctypes in intranet

### DIFF
--- a/site/components/src/main/java/uk/nhs/digital/intranet/components/AnnouncementComponent.java
+++ b/site/components/src/main/java/uk/nhs/digital/intranet/components/AnnouncementComponent.java
@@ -1,0 +1,13 @@
+package uk.nhs.digital.intranet.components;
+
+import org.hippoecm.hst.core.component.HstComponentException;
+import org.hippoecm.hst.core.component.HstRequest;
+import org.hippoecm.hst.core.component.HstResponse;
+import uk.nhs.digital.common.components.BaseGaContentComponent;
+
+public class AnnouncementComponent extends BaseGaContentComponent {
+
+    @Override
+    public void doBeforeRender(HstRequest request, HstResponse response) throws HstComponentException {
+    }
+}
--- a/repository-data/application/src/main/resources/hcm-config/configuration/queries/templates/new-announcement-document.yaml
+++ b/repository-data/application/src/main/resources/hcm-config/configuration/queries/templates/new-announcement-document.yaml
@@ -1,0 +1,27 @@
+---
+definitions:
+  config:
+    /hippo:configuration/hippo:queries/hippo:templates/new-announcement-document:
+      hippostd:icon: adddocument_ico
+      hippostd:modify:
+      - ./_name
+      - $name
+      - ./hippotranslation:locale
+      - $inherited
+      - ./hippotranslation:id
+      - $uuid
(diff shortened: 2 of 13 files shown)


site/components/src/main/java/uk/nhs/digital/intranet/components/AnnouncementComponent.java
```diff
@@ -0,0 +1,23 @@
...
+
+    @Override
+    public void doBeforeRender(HstRequest request, HstResponse response) throws HstComponentException {
+        super.doBeforeRender(request, response);
+        HstRequestContext requestContext = RequestContextProvider.get();
+        HippoBean contentBean = requestContext.getContentBean();
+
+        if (contentBean != null) {
+            request.setAttribute("document", contentBean);
+        }
```
mohammed-hasan-valtech, line 21: Can you just use the BaseGaComponent ? Does the same thing + also allows full ga tracking in html fields

mohammed-hasan-valtech, line 21: Can you just use the BaseGaComponent ? Does the same thing + also allows full ga tracking in html fields

alexandru-mitcu-valtech: Good spot

site/components/src/main/java/uk/nhs/digital/intranet/components/AnnouncementComponent.java
```diff
@@ -0,0 +1,13 @@
+package uk.nhs.digital.intranet.components;
+
+import org.hippoecm.hst.core.component.HstComponentException;
+import org.hippoecm.hst.core.component.HstRequest;
+import org.hippoecm.hst.core.component.HstResponse;
+import uk.nhs.digital.common.components.BaseGaContentComponent;
+
+public class AnnouncementComponent extends BaseGaContentComponent {
```
mohammed-hasan-valtech, line 8: Can you directly use BaseGaContentComponent in the hst:page? it removes the extra component being instantiated

alexandru-mitcu-valtech: Oooh, that's what you meant. It shouldn't be any problem to re-use the same Java component for multiple hst:pages, correct?

alexandru-mitcu-valtech: On the other hand, I would argue against this on the basis that this is not something done in the codebase (or at least I didn't notice it). My view on this is that every page should have its own component, and that the `BaseGaContentComponent` should only be extended, not directly used. The same as with `BaseHstDocument`.

mohammed-hasan-valtech: I believe it is recommended to reuse these base components for multiple simple doctypes, but its not a big deal either way, I'll approve.

PR #922, opened 2020-04-21
[DW-1471] Adding tableau compound, using JS to render tableau Viz

ShaneAhern-IRE:


repository-data/webfiles/src/main/resources/site/freemarker/common/macro/sections/tableauSection.ftl
```diff
@@ -0,0 +1,34 @@
...
+            <script type="text/javascript"
+                    src="https://public.tableau.com/javascripts/api/tableau-2.min.js"></script>
+        </@hst.headContribution>
+
+        <#assign divId = "tableau-${index}"/>
+
+        <div id="${divId}"></div>
+
+        <script type="text/javascript">
+            window.onload(initViz());
```
GeoffHayward, line 18: The `window.loaded` loaded event is limited to just one per page. This therefore could conflict with other scripts. You should use `window.addEventListener('load', initViz);` instead as this event is not limited to one.

ShaneAhern-IRE: Added eventListener to invoke the function trigger

PR #928, opened 2020-04-24
[DW-1476] Task page spacing & chapter link check

josephcyrus-valtech:


repository-data/application/src/main/resources/hcm-content/content/documents/administration/intranet/task.yaml
```diff
@@ -0,0 +1,73 @@
+---
```
alexandru-mitcu-valtech, line 1: Is this a `Task` resource bundle document?

mohammed-hasan-valtech: yes, i've just added the existing Task resource bundles so they can be used when devleoping/testing. They won't be bootstrapped into live as I've already had them manually imported there

PR #944, opened 2020-05-06
[DW-1458] Add Navigation item UI

JodiWarren:
- Add markup
- Style tiles
- Restrict number of links to 1


repository-data/webfiles/src/main/resources/site/scss/components/_navigation-tiles.scss
```diff
@@ -0,0 +1,138 @@
+.navigation-section {}
+
+.navigation-section__introduction {
+    margin-bottom: 20px;
+}
+
+.navigation-section--column .navigation-section__tiles {
+    @include media(desktop) {
+        display: flex;
```
GeoffHayward, line 9: Have you validated this works in IE 11? It seems there is partial support for `display: flex` according to 'Can I Use' https://caniuse.com/#search=display%3A%20flex

repository-data/webfiles/src/main/resources/site/scss/components/_navigation-tiles.scss
```diff
@@ -0,0 +1,138 @@
...
+        max-width: 528px;
+        text-align: left;
+    }
+}
+
+.navigation-tile__title {
+    font-size: 22px;
+    line-height: 1.32;
+    font-weight: bold;
+    color: currentColor;
```
GeoffHayward, line 93: I am sure it's fine, but please can you talk me through this? I had thought SCSS variables begin with a '$' dollar sign.

repository-data/webfiles/src/main/resources/site/scss/components/_navigation-tiles.scss
JodiWarren: I actually wrote this some time ago, so I should probably circle back around and test it again. I use Flexbox in IE all the time, and often rely on https://github.com/philipwalton/flexbugs to solve small implementation-specific issues.

I'm 99% sure that `display: flex` works on IE11, and that the results on caniuse are due to a confusing table on MDN (where it gets it's source data from). But will double check!

repository-data/webfiles/src/main/resources/site/scss/components/_navigation-tiles.scss
JodiWarren: Today's your lucky day! This is a great CSS feature.

`currentColor` has had full browser support since IE9: https://caniuse.com/#feat=currentcolor. It's a way of using the currently defined value of `color` for other properties.

See https://codepen.io/jodi/pen/VwvQRYW for a silly example

repository-data/webfiles/src/main/resources/site/scss/components/_navigation-tiles.scss
JodiWarren: Here's a screenshot of it running in IE11, tested via Browserstack

bs_win10_IE_11 0

repository-data/webfiles/src/main/resources/site/scss/components/_navigation-tiles.scss
GeoffHayward: That is a great feature 👍 Thanks for letting me know.

PR #952, opened 2020-05-19
[DW-1385] Upgrading project to 13.4.2 

ShaneAhern-IRE:


pom.xml
```diff
@@ -627,7 +627,15 @@
                 <configuration>
                     <rules>
                         <reactorModuleConvergence/>
-                        <requireUpperBoundDeps/>
+                        <requireUpperBoundDeps>
+                            <excludes>
```
ShaneAhern-IRE, line 631: Required to compile. Bloomreach parent project cannot guarantee that they have the upper bound dependency.

site/components/src/test/java/uk/nhs/digital/intranet/provider/MicrosoftGraphAuthorizationProviderTest.java
```diff
@@ -137,4 +138,11 @@ public void throwsAuthorizationExceptionIfGraphApiThrowsHttpExceptionOnRefreshAc
         headers.setContentType(MediaType.APPLICATION_FORM_URLENCODED);
         return new HttpEntity<>(map, headers);
     }
+
+    private boolean timeWithin5Seconds(AccessToken accessToken) {
```
ShaneAhern-IRE, line 142: Added more flexibility to time check ensuring that the token calculated is within 5 seconds of the correct time rather than exact second which was causing unit tests to fail intermittently.

PR #958, opened 2020-06-02
[DW-1500] Improvise dynamic charts

GeoffHayward:
- Adds accessibility
- Adds a download link
- Adds caching via CRISP

The CSV is pre-fetched and then  downloaded with the web 
page in base 64. This helps the download work and negated 
the need for a loading wait.


ziemsky:
All is looking good and reasonable as far as I can tell.

Nice use of CRISP API and the request interceptor - and a good example for me to steal for one of my soon-to-be-worked-on stories :wink:

PR #967, opened 2020-06-12
[DW-1540] Fixes Statistics load error

GeoffHayward:
Also fixes dynamic chart and refactors both
such that they share the fix.
Also refactors some Freemarker utils into a
utils package.
Also adds RAW text response to statistic and 
removes the unused 'dataType' field.


site/components/src/main/java/uk/nhs/digital/freemarker/utils/package-info.java
```diff
@@ -0,0 +1,8 @@
+/**
+ * A collection of generic helper Freemarker functions.
+ *
+ * Freemarker template usage example:
+ *  <#assign inArray="uk.nhs.digital.freemarker.utils.InArray"?new() />
+ *  <#if inArray(options.status?lower_case, ['complete'])>
+ */
+package uk.nhs.digital.freemarker.utils;
```
ziemsky, line 8: :heart_eyes:

:+1: for package documentation - I don't see it (or apply :disappointed: myself) often enough.

PR #968, opened 2020-06-15
[ADZ-127] CRISP API for Apigee access

ziemsky:
Incremental/preliminary Pull Request against a long-running feature branch `pds_api_spec_epic`, in order to avoid one, massive PR later.


cms/src/main/resources/META-INF/hst-assembly/addon/crisp/overrides/custom-resource-resolvers.xml
```diff
@@ -0,0 +1,48 @@
...
+          class="org.onehippo.cms7.crisp.core.resource.jackson.SimpleJacksonRestTemplateResourceResolver">
+        <property name="cacheEnabled" value="false"/>
+        <property name="restTemplate">
+            <bean class="org.springframework.security.oauth2.client.OAuth2RestTemplate">
+                <constructor-arg>
+                    <bean class="org.springframework.security.oauth2.client.token.grant.password.ResourceOwnerPasswordResourceDetails">
+                        <property name="clientAuthenticationScheme" value="form"/>
+                        <property name="grantType" value="password"/>
+                        <property name="accessTokenUri" value="${devzone.apigee.oauth.token.url}"/>
+                        <property name="username" value="${DEVZONE_APIGEE_OAUTH_USERNAME}"/>
```
GeoffHayward, line 17: Are the new environment configurations documented?

pom.xml
```diff
@@ -69,6 +69,9 @@
 
         <hippo.cms.locales></hippo.cms.locales>
 
+        <!-- Keep in line with BloomReach version - see <parent><version> above -->
+        <hippo.version>13.4.2</hippo.version>
```
GeoffHayward, line 73: Is this value already available? I worry about it going out of sync.

cms/src/main/resources/META-INF/hst-assembly/addon/crisp/overrides/custom-resource-resolvers.xml
ziemsky: They will be towards the end of the epic's implementation. I have a ticket to raise about that in my todo.

pom.xml
ziemsky: Good question. I worried about the same thing too, but I couldn't figaure out any better way.

What happens is that we need dependency `org.onehippo.cms7:hippo-addon-crisp-mock` in `cms/pom.xml` and, in order to make it available, I added it to main POM's `dependencyManagement` first.

Here (i.e. in the main POM), I was trying to specify this version using `${project.parent.version}` to keep it automatically in sync with the actual BloomReach version.

Unfortunately, this variable gets resolved in the context of the '_child_' pom, where it would receive value of `dev-SNAPSHOT`, because the `parent` part evaluates there to the main POM. This is described in [Maven's documentation](http://maven.apache.org/guides/introduction/introduction-to-the-pom.html#project-interpolation-and-variables):
> One factor to note is that these variables are processed after inheritance as outlined above. This means that if a parent project uses a variable, then its definition in the child, not the parent, will be the one eventually used.

Sadly, neither `${project.parent.parent.version}` nor `${root.project.version} are a thing (I tried) :disappointed:. No amount of googling gave me a better solution; what I could find, though is frustrations of others about the lack of good support for such a feature ([e.g. here](https://stackoverflow.com/a/40363629)).

We already have a precedence for this, it seems in the `<essentials.version>` property. It now occurs to me that it will be better to merge the two properties into one. A commit is shortly on its way.

PR #975, opened 2020-06-25
[DW-1516] Fixes emphasis box headings

GeoffHayward:


repository-data/webfiles/src/main/resources/site/freemarker/common/macro/sections/emphasisBox.ftl
```diff
@@ -38,7 +38,7 @@
 
         <div class="emphasis-box__content">
             <#if section.heading?has_content && slug??>
-                <h3 id="${slugify(slug)}" data-uipath="website.contentblock.emphasis.heading">${section.heading}</h3>
+                <strong class="emphasis-box__heading" id="${slugify(slug)}" data-uipath="website.contentblock.emphasis.heading">${section.heading}</strong>
```
JodiWarren: By removing the header tag from this, it means that the user won't necessarily know that this line is a heading line, and won't be able to navigate to it as a heading/landmark. I just tested it with Voiceover, and it just reads `<strong>` text as if it was any other text. If this is correct, then please dismiss this review. If not, you may be able to add it as a title with an `aria-title` attribute on the containing `div`, or perhaps wrap it in an `aside` and keep the heading tag.

PR #988, opened 2020-07-15
[ADZ-55] API Specifications - Try it now

ziemsky:
Another incremental pull request against a feature branch (rather than `master`) to avoid one huge PR later.


repository-data/webfiles/src/main/resources/site/freemarker/common/apispecification.ftl
```diff
@@ -1,19 +1,41 @@
...
+            <@documentHeader document 'general'></@documentHeader>
+
+            <style type="text/css">
+                .ctabtn--nhs-digital-button--try-it-now { float: right; }
+            </style>
+
+            <div class="grid-wrapper grid-wrapper--article">
+                <div class="grid-row">
+
+                    ${document.html?no_esc}
```
GeoffHayward, line 28: Re: conversation on supplying a value to a mustache template's href from this later context, you could add a placeholder in mustache and then do a string replace here. For anything more complicated, such as replacing more than one value, you could feed `document.html` to an implementation of a Freemarker TemplateMethodModelEx. Then print the returned object from your TemplateMethodModelEx.

ziemsky: THAT is a cool idea. Thanks, Geoff.

Admittedly, it means that the entire (substantial) `document.html` payload will have to be processed on each request but there are many documents with large-is payload that we process with Freemarker anyway, so it shouldn't really be a concern.

PR #1015, opened 2020-08-14
[ADZ-283] Display request body content type

ziemsky:


ziemsky:
Code-wise, the force-pushed commit `18212c1` is identical to `f0b9843` and was only created to correct the story reference in the comment. Comment of the `f0b9843` commit (and the branch name) incorrectly referenced story ADZ-286 whilst the change was actually for [ADZ-283](https://nhsd-jira.digital.nhs.uk/browse/ADZ-283).

PR #1049, opened 2020-09-07
[ADZ-285] Response Body Examples

ziemsky:


monika10793 (approved):

ziemsky:
Commit `e1636b1`:
* Updates the branch with the latest changes from `master`.
* Adds support for an edge case missed from the original commit: `examples` from Media Type Object override `example` from the object's schema. The change boils down to the addition of `{{^examples}} ... {{/examples}}` tags in `body_param.mustache` and updating test data files.

PR #1074, opened 2020-10-23
[DW-000] Fix for missing labels

GeoffHayward:
Add the new content as it's in a new file.


JodiWarren (approved):
I don't fully understand the implementation purpose but it appears to be syntactically correct

PR #1076, opened 2020-10-27
[ADZ-531] Hide headers when sections empty

monika10793:
Hide headers when sections empty


cms/src/main/java/uk/nhs/digital/apispecs/swagger/IsAnyTrueHelper.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public static final IsAnyTrueHelper INSTANCE = new IsAnyTrueHelper();
+
+    public static final String NAME = "isAnyTrue";
+
+    @Override public Object apply(final String items, final Options options) throws IOException {
+
+        Validate.notEmpty(options.params,"Exception in IsAnyTrueHelper");
+        Validate.noNullElements(options.params,"Exception in IsAnyTrueHelper");
+        Validate.allElementsOfType(Arrays.asList(options.params), Boolean.class,"Exception in IsAnyTrueHelper");
```
ziemsky, line 21: Stack trace in the logs will already inform the readers that an exception was thrown and from which class.

Therefore, in their current form, these messages do not add any value. Instead, they should be communicating which aspect of validation has failed.

cms/src/main/java/uk/nhs/digital/apispecs/swagger/IsAnyTrueHelper.java
```diff
@@ -0,0 +1,44 @@
...
+        if (hasRequestParams(options)) {
+            buffer.append(options.fn());
+        } else {
+            buffer.append(options.inverse());
+        }
+
+        return buffer;
+    }
+
+    private boolean hasRequestParams(final Options options) {
```
ziemsky, line 39: This method name is misleading, as this is a quite generic helper which may be used in many contexts, other than Request section.

What does this method assert?

cms/src/main/java/uk/nhs/digital/apispecs/swagger/IsAnyTrueHelper.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public static final String NAME = "isAnyTrue";
+
+    @Override public Object apply(final String items, final Options options) throws IOException {
+
+        Validate.notEmpty(options.params,"Exception in IsAnyTrueHelper");
+        Validate.noNullElements(options.params,"Exception in IsAnyTrueHelper");
+        Validate.allElementsOfType(Arrays.asList(options.params), Boolean.class,"Exception in IsAnyTrueHelper");
+
+        if (!(options.params.length > 0)) {
```
ziemsky, line 23: Doesn't the `Validate` class provide a suitable method to assert this?

cms/src/main/resources/api-specification/codegen-templates/index.mustache
```diff
@@ -70,7 +70,9 @@
                                             <p></p>
                                             <br/>
 
-                                            <h2>Request</h2>
+                                            {{#isAnyTrue Boolean hasPathParams hasHeaderParams hasBodyParam hasFormParams hasQueryParams}}
```
ziemsky, line 73: I'm afraid that the use of `Boolean` here is incorrect; admittedly, it probably doesn't break anything but it is misleading. 

If I remember what Handlebars documentation says, the first argument either represents the current context or even is optional.

It would be best to remove this argument if it's indeed optional, or replace it with something more meaningful.

cms/src/test/java/uk/nhs/digital/apispecs/IsAnyTrueHelperTest.java
```diff
@@ -0,0 +1,126 @@
...
+        assertThat("Returns Options.Buffer",
+            buffer,
+            instanceOf(Options.Buffer.class)
+        );
+
+        final Options.Buffer actualBuffer = (Options.Buffer) buffer;
+
+        then(actualBuffer).should().append(TEMPLATE_CONTENT_FROM_THE_POSITIVE_BLOCK);
+    }
+
```
ziemsky, line 53: We need a test case asserting that the 'negative' block is rendered.

cms/src/test/java/uk/nhs/digital/apispecs/IsAnyTrueHelperTest.java
```diff
@@ -0,0 +1,126 @@
...
+            isAnyTrueHelper.apply(item, options);
+
+        } catch (Exception e) {
+
+            // then
+            assertThat("Exception in execution","Exception in IsAnyTrueHelper", is(e.getMessage()));
+        }
+    }
+
+    class MockOptions extends Options {
```
ziemsky, line 105: This class is actually a 'stub' rather than a 'mock'. It would be good to reflect this fact in its name.

cms/src/test/java/uk/nhs/digital/apispecs/SwaggerCodeGenOpenApiSpecificationJsonToHtmlConverterTest.java
```diff
@@ -72,6 +72,33 @@ public void getHtmlForSpec_throwsException_onCodeGenFailure() {
         // expectations set in 'given' are satisfied
     }
 
+    @Test
+    public void hideRequestHeaderWhenSectionIsEmpty() {
```
ziemsky, line 76: The test data files used here are unnecessarily large for testing a small test case. This makes them hard to maintain and understand, and their size contributes to log limit being exceeded on Travis CI in the case of test failure, as their content is logged in the entirety.

The files should be cut down to the minimum necessary content required to exercise this test case.

Either the files should be cut

cms/src/test/java/uk/nhs/digital/apispecs/SwaggerCodeGenOpenApiSpecificationJsonToHtmlConverterTest.java
```diff
@@ -72,6 +72,33 @@ public void getHtmlForSpec_throwsException_onCodeGenFailure() {
...
+        // then
+        final String expectedSpecHtml = customisedSwaggerCodeGenGeneratedSpecificationHtmlForEmptyReqSection();
+
+        assertThat(
+            "Specification HTML has been generated using customised Swagger CodeGen v3",
+            actualSpecHtml,
+            is(expectedSpecHtml)
+        );
+    }
+
```
ziemsky, line 93: Given that we have the above test case, we need to demonstrate the same for Response and Schema sections.

Given that the Schema is rendered through the `body_params` template, the `BodyRenderTest` class is probably a better fit for testing Schema section.

ziemsky (changes requested):

cms/src/test/resources/test-data/api-specifications/SwaggerCodeGenOpenApiSpecificationJsonToHtmlConverterTest/customised-codegen-v3-generated-spec.html
```diff
@@ -1235,14 +1238,104 @@ <h3>Query parameters</h3>
...
+                                                    <h2>Response</h2>
+
                                                 <h3>HTTP status: 200</h3>
                                                 <div>A completed search, containing zero, one, or many matching patients.</div>
 
 
                                                         <h4>Body</h4>
                                                     
                                                     
+                                                            <h5 class="body__contenttype">
```
ziemsky, line 1250: I may be missing something but it feels surprising that this large block of content would be added here.

Is this definitely correct?

cms/src/test/resources/test-data/api-specifications/SwaggerCodeGenOpenApiSpecificationJsonToHtmlConverterTest/customised-codegen-v3-generated-spec.html
```diff
@@ -2133,14 +2227,103 @@ <h3>Body</h3>
                                                     <div class="httpparams__required">Required</div>
                                                         
                                                         
+                                                                <h5 class="body__contenttype">
```
ziemsky, line 2230: Similarly to the previous comment: is it definitely correct that this block of content is added here?

cms/src/test/resources/test-data/api-specifications/ApiSpecSyncFromApigeeJobIntegrationTest/customised-codegen-v3-generated-spec.html
```diff
@@ -1235,14 +1238,104 @@ <h3>Query parameters</h3>
...
+                                                    <h2>Response</h2>
+
                                                 <h3>HTTP status: 200</h3>
                                                 <div>A completed search, containing zero, one, or many matching patients.</div>
 
 
                                                         <h4>Body</h4>
                                                     
                                                     
+                                                            <h5 class="body__contenttype">
```
ziemsky, line 1250: Similarly to the previous comments: is it definitely correct that this block of content is added here?

cms/src/test/resources/test-data/api-specifications/ApiSpecSyncFromApigeeJobIntegrationTest/customised-codegen-v3-generated-spec.html
```diff
@@ -2133,14 +2227,103 @@ <h3>Body</h3>
                                                     <div class="httpparams__required">Required</div>
                                                         
                                                         
+                                                                <h5 class="body__contenttype">
```
ziemsky, line 2230: Similarly to the previous comment: is it definitely correct that this block of content is added here?

ziemsky (changes requested):

cms/src/test/java/uk/nhs/digital/apispecs/IsAnyTrueHelperTest.java
```diff
@@ -0,0 +1,126 @@
...
+    @Mock private Options.Buffer buffer;
+    private String item = "some-item";
+
+    @Before
+    public void setUp() throws Exception {
+        initMocks(this);
+    }
+
+    @Test
+    public void rendersBlockWhenCollectionHasAtLeastOneItem() throws IOException {
```
ziemsky, line 35: This probably was meant to say 'has at least one _true_ item'?

ziemsky (changes requested):

cms/src/test/java/uk/nhs/digital/apispecs/IsAnyTrueHelperTest.java
```diff
@@ -0,0 +1,126 @@
...
+
+        } catch (Exception e) {
+
+            // then
+            assertThat("Exception in execution","Exception in IsAnyTrueHelper", is(e.getMessage()));
+        }
+    }
+
+    @Test
+    public void throwExceptionWhenCollectionIsNotBoolean() throws IOException {
```
ziemsky, line 89: Will this test fail when the helper throws no exception?

cms/src/test/java/uk/nhs/digital/apispecs/IsAnyTrueHelperTest.java
```diff
@@ -0,0 +1,126 @@
...
+
+        } catch (Exception e) {
+
+            // then
+            assertThat("Exception in execution","Exception in IsAnyTrueHelper", is(e.getMessage()));
+        }
+    }
+
+    @Test
+    public void throwExceptionWhenCollectionIsNull() throws IOException {
```
ziemsky, line 72: Will this test fail when the helper throws no exception?

cms/src/test/java/uk/nhs/digital/apispecs/IsAnyTrueHelperTest.java
```diff
@@ -0,0 +1,126 @@
...
+            instanceOf(Options.Buffer.class)
+        );
+
+        final Options.Buffer actualBuffer = (Options.Buffer) buffer;
+
+        then(actualBuffer).should().append(TEMPLATE_CONTENT_FROM_THE_POSITIVE_BLOCK);
+    }
+
+    @Test
+    public void throwExceptionWhenCollectionIsEmpty() throws IOException {
```
ziemsky, line 55: Will this test fail when the helper throws no exception?

ziemsky (changes requested):

cms/src/main/java/uk/nhs/digital/apispecs/swagger/IsAnyTrueHelper.java
```diff
@@ -0,0 +1,39 @@
...
+
+public class IsAnyTrueHelper implements Helper<String> {
+
+    public static final IsAnyTrueHelper INSTANCE = new IsAnyTrueHelper();
+
+    public static final String NAME = "isAnyTrue";
+
+    @Override public Object apply(final String items, final Options options) throws IOException {
+
+        Validate.notEmpty(options.params,"One or more input parameters");
```
ziemsky, line 19: I'm afraid this version of the message still doesn't communicate what the failure was.

When I make this validation fail, what I'm getting in the logs is `java.lang.IllegalArgumentException: One or more input parameters` and still have to analyse source code to figure out what the error was about.

This validation requires that at least one parameter is provided.

Why not communicate this explicitly in the message, saying something like 'At least one parameter is required but none was provided.'?

cms/src/main/java/uk/nhs/digital/apispecs/swagger/IsAnyTrueHelper.java
```diff
@@ -0,0 +1,39 @@
...
+        Validate.noNullElements(options.params,"One or more input parameters");
```
ziemsky, line 20: Similar [comment to the previous one](https://github.com/NHS-digital-website/hippo/pull/1076/files#diff-07955cf23a101e3697055922bbf7714e72910000b41fa2eaeb94da5af7de09d1R19); as it is now the message is doesn't add much value.

cms/src/main/java/uk/nhs/digital/apispecs/swagger/IsAnyTrueHelper.java
```diff
@@ -0,0 +1,39 @@
...
+        Validate.allElementsOfType(Arrays.asList(options.params), Boolean.class,"Required parameter type is Boolean");
```
ziemsky, line 21: I'm afraid that in the current form this message is confusing and, potentially, misleading. 

This validation requires that all arguments are of type Boolean and fails if even one is not, and the error message should state it explicitly if we are using it at all.

As is, it could be easily misinterpreted as saying 'the method requires a non-boolean parameter but boolean was provided.

Yes, looking at the stack trace to see which validation method the exception came from does add clarifying context but if we require people to look at the stack trace to understand the problem, why use error messages at all?

cms/src/main/java/uk/nhs/digital/apispecs/swagger/IsAnyTrueHelper.java
```diff
@@ -0,0 +1,39 @@
...
+        if (hasParams(options)) {
+            buffer.append(options.fn());
+        } else {
+            buffer.append(options.inverse());
+        }
+
+        return buffer;
+    }
+
+    private boolean hasParams(final Options options) {
```
ziemsky, line 34: This method name is still misleading.

The method checks whether any of the options' parameters is true. The method's name, however, misleadingly communicates something different. As is, it says 'it returns true if options has any parameters at all'.

cms/src/test/java/uk/nhs/digital/apispecs/IsAnyTrueHelperTest.java
```diff
@@ -0,0 +1,155 @@
...
+        } catch (Exception e) {
+
+            // then
+            assertThat("Exception in execution",
+                e.getMessage(),
+                containsString("Required parameter type is Boolean")
+            );
+        }
+    }
+
```
ziemsky, line 133: Since we are verifying each validation with dedicated test case, then I think that it makes sense to have one for `Validate.noNullElements`.

cms/src/test/resources/test-data/api-specifications/SwaggerCodeGenOpenApiSpecificationJsonToHtmlConverterTest/openapi-v3-spec-empty-res-sec.json
```diff
@@ -0,0 +1,4019 @@
+{
+    "openapi": "3.0.0",
+    "info": {
+        "version": "v1.2.29-beta",
+        "title": "Personal Demographics Service (FHIR) API",
```
ziemsky, line 5: I'm afraid this file still contains a lot of content that does nothing to illustrate the test but, instead, obscures the content that is relevant. The same applies to the corresponding HTML file.

For example the `info.description` field has a very large content but has zero impact on the test; no only its content is irrelevant, but even the presence of the field itself is redundant. All we need is a minimal, necessary content that surrounds the one that the test actually cares about.

Admittedly, we do use an almost complete PDS spec content in the `getHtmlForSpec_convertsProvidedOpenApiJsonToHtml` but this is a 'catch all', high level tests, written in the time when we had very little understanding of the specifications' content and PDS was our exemplar. Even that content is too big and full of redundancies and there is a tech debt ticket to clean it up.

For the test cases that are small and focused (like the one that uses the current file), the content should be in order of tens of lines, not five thousands as it's very hard to see what's relevant. File `exampleOnMediaTypeObjectNoneOnSchema.json` is a good example of a scale of small test data file that focuses only relevant data; naturally the current file needs to be slightly bigger but not by much.

cms/src/test/resources/test-data/api-specifications/SwaggerCodeGenOpenApiSpecificationJsonToHtmlConverterTest/openapi-v3-spec-empty-req-sec.json
```diff
@@ -0,0 +1,5263 @@
+{
+    "openapi": "3.0.0",
+    "info": {
```
ziemsky, line 3: This file - as well as the corresponding HTML one - still looks way too big.

See comment to file [openapi-v3-spec-empty-req-sec.json](https://github.com/NHS-digital-website/hippo/pull/1076/files#diff-0b473cb70a9920aa524c94df9f7d7f97609d6bafb6ee5c73251e88f053927bd2).

ziemsky (changes requested):

cms/src/main/java/uk/nhs/digital/apispecs/swagger/IsAnyTrueHelper.java
```diff
@@ -0,0 +1,39 @@
...
+        return Arrays.stream(Optional.ofNullable(options).get().params)
```
ziemsky, line 35: The use of `Optional` is redundant here.

We receive options from Handlebars and we can rely on the framework that it can do the right thing. Besides, there are calls to member field of `options` prior to this method that would fail much earlier before this method would be called. Also, as written, the call to `get()` would fail if `options` were `null`.

cms/src/main/java/uk/nhs/digital/apispecs/swagger/IsAnyTrueHelper.java
monika10793: Fixed

cms/src/main/java/uk/nhs/digital/apispecs/swagger/IsAnyTrueHelper.java
monika10793: Fixed

cms/src/main/java/uk/nhs/digital/apispecs/swagger/IsAnyTrueHelper.java
monika10793: Fixed

cms/src/main/resources/api-specification/codegen-templates/index.mustache
monika10793: Fixed

cms/src/test/java/uk/nhs/digital/apispecs/SwaggerCodeGenOpenApiSpecificationJsonToHtmlConverterTest.java
monika10793: Fixed

cms/src/test/java/uk/nhs/digital/apispecs/SwaggerCodeGenOpenApiSpecificationJsonToHtmlConverterTest.java
monika10793: Fixed

cms/src/test/resources/test-data/api-specifications/SwaggerCodeGenOpenApiSpecificationJsonToHtmlConverterTest/customised-codegen-v3-generated-spec.html
monika10793: Fixed

cms/src/test/resources/test-data/api-specifications/SwaggerCodeGenOpenApiSpecificationJsonToHtmlConverterTest/customised-codegen-v3-generated-spec.html
monika10793: Fixed

cms/src/test/resources/test-data/api-specifications/ApiSpecSyncFromApigeeJobIntegrationTest/customised-codegen-v3-generated-spec.html
monika10793: Fixed

cms/src/test/resources/test-data/api-specifications/ApiSpecSyncFromApigeeJobIntegrationTest/customised-codegen-v3-generated-spec.html
monika10793: Fixed

cms/src/test/java/uk/nhs/digital/apispecs/IsAnyTrueHelperTest.java
monika10793: Fixed

cms/src/test/java/uk/nhs/digital/apispecs/IsAnyTrueHelperTest.java
monika10793: Fixed

cms/src/test/java/uk/nhs/digital/apispecs/IsAnyTrueHelperTest.java
monika10793: Fixed

cms/src/test/java/uk/nhs/digital/apispecs/IsAnyTrueHelperTest.java
```diff
@@ -0,0 +1,146 @@
...
+        Boolean[] params = {false,false,true,false,false};
+        StubOptions options = new StubOptions(params);
+
+        // when
+        final Object buffer = isAnyTrueHelper.apply(item, options);
+
+        // then
+        assertThat("Returns Options.Buffer",
+            buffer,
+            instanceOf(Options.Buffer.class)
```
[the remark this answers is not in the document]
monika10793: Fixed

cms/src/test/java/uk/nhs/digital/apispecs/IsAnyTrueHelperTest.java
```diff
@@ -0,0 +1,146 @@
...
+        );
+
+        final Options.Buffer actualBuffer = (Options.Buffer) buffer;
```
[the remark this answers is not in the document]
monika10793: Fixed

cms/src/test/resources/test-data/api-specifications/SwaggerCodeGenOpenApiSpecificationJsonToHtmlConverterTest/openapi-v3-specification.json
```diff
@@ -248,6 +248,51 @@
...
+                                    }
+                                }
+                            },
+                            "application/render-example-from-schema-where-none-present-on-object": {
+                                "schema": {
+                                    "example": {
+                                        "example-value": "example on schema - should render as pretty-printed JSON"
+                                    }
+                                }
+                            },
```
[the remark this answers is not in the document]
monika10793: Fixed

cms/src/test/java/uk/nhs/digital/apispecs/swagger/BodyRenderingTest.java
```diff
@@ -162,6 +162,40 @@ public void rendersCompleteContentTypeSections_withAllComponentsInTheRightOrder(
...
+    public void rendersSchemaWhenMediaTypesSchemaIsPresent() throws IOException {
+
+        // given
+        final String bodyJson = fromFile("exampleOnSchemaNoneOnMediaTypeObject.json");
+
+        // when
+        final String actualHtml = renderFor(bodyJson);
+
+        // then
+        assertThat("Does not render example from MediaTypeObject's Schema.",
```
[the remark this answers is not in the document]
monika10793: Fixed

cms/src/test/java/uk/nhs/digital/apispecs/swagger/BodyRenderingTest.java
```diff
@@ -162,6 +162,40 @@ public void rendersCompleteContentTypeSections_withAllComponentsInTheRightOrder(
...
+    public void doesNotRendersSchemaWhenMediaTypesSchemaIsAbsent() throws IOException {
+
+        // given
+        final String bodyJson = fromFile("exampleOnMediaTypeObjectNoneOnSchema.json");
+
+        // when
+        final String actualHtml = renderFor(bodyJson);
+
+        // then
+        assertThat("Does not render example from MediaTypeObject's Schema.",
```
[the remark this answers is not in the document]
monika10793: Fixed

cms/src/main/java/uk/nhs/digital/apispecs/swagger/IsAnyTrueHelper.java
monika10793: Fixed

cms/src/main/java/uk/nhs/digital/apispecs/swagger/IsAnyTrueHelper.java
monika10793: Fixed

cms/src/test/java/uk/nhs/digital/apispecs/IsAnyTrueHelperTest.java
monika10793: Fixed

cms/src/test/resources/test-data/api-specifications/SwaggerCodeGenOpenApiSpecificationJsonToHtmlConverterTest/openapi-v3-spec-empty-res-sec.json
monika10793: Fixed

cms/src/test/resources/test-data/api-specifications/SwaggerCodeGenOpenApiSpecificationJsonToHtmlConverterTest/openapi-v3-spec-empty-req-sec.json
monika10793: Fixed

cms/src/main/java/uk/nhs/digital/apispecs/swagger/IsAnyTrueHelper.java
monika10793: Fixed

cms/src/test/java/uk/nhs/digital/apispecs/IsAnyTrueHelperTest.java
ziemsky: I'm afraid that if I make the helper throw no exception, this test still passes - and yet, it was meant to verify that we throw validation exception with a specific message.

Indeed, if I _only_ comment out `Validate.allElementsOfType` it does fail but it's not because _validation_ exception was thrown.

This test needs to assert that the exception thrown is of the type thrown by the `Validate` method it verifies  - otherwise it's not checking this specific validation. Also, it should fail when absolutely no exception is thrown from the method under test.

ziemsky (changes requested):

cms/src/test/java/uk/nhs/digital/apispecs/IsAnyTrueHelperTest.java
```diff
@@ -0,0 +1,150 @@
...
+        } catch (Exception e) {
+
+            // then
+            assertThat("Exception in execution",
+                e.getMessage(),
+                containsString("At least one parameter is required but none was provided.")
+            );
+        }
+    }
+
```
ziemsky, line 108: This test case verifies that a validation exception is thrown when the value of `options.params` is `null`.

Which test case verifies that a validation exception is thrown when either of its elements is `null`? In other words, which test case targets `Validation.noNullElements`?

cms/src/test/java/uk/nhs/digital/apispecs/IsAnyTrueHelperTest.java
monika10793: Fixed

cms/src/test/resources/test-data/api-specifications/SwaggerCodeGenOpenApiSpecificationJsonToHtmlConverterTest/openapi-v3-spec-empty-req-sec.json
monika10793: Fixed

cms/src/test/java/uk/nhs/digital/apispecs/IsAnyTrueHelperTest.java
monika10793: fixed

cms/src/main/java/uk/nhs/digital/apispecs/swagger/IsAnyTrueHelper.java
```diff
@@ -0,0 +1,38 @@
...
+public class IsAnyTrueHelper implements Helper<String> {
+
+    public static final IsAnyTrueHelper INSTANCE = new IsAnyTrueHelper();
+
+    public static final String NAME = "isAnyTrue";
+
+    @Override public Object apply(final String items, final Options options) throws IOException {
+
+        Validate.notEmpty(options.params,"At least one parameter is required but none was provided.");
+        Validate.noNullElements(options.params,"At least one parameter is required but none was provided.");
```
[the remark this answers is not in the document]
monika10793: noNullElements check is redundant now.

Validate.notEmpty() checks for null

cms/src/main/java/uk/nhs/digital/apispecs/swagger/IsAnyTrueHelper.java
```diff
@@ -0,0 +1,38 @@
...
+        Validate.allElementsOfType(Arrays.asList(options.params), Boolean.class,"The method requires a boolean parameter but non-boolean was provided.");
```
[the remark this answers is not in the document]
monika10793: noNullElements check is redundant now.

Validate.notEmpty() checks for null

cms/src/test/java/uk/nhs/digital/apispecs/IsAnyTrueHelperTest.java
monika10793: following test case check for when either of element in collection is null.

throwExceptionWhenAtLeastOneIsNull()

cms/src/main/java/uk/nhs/digital/apispecs/swagger/IsAnyTrueHelper.java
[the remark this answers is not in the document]
ziemsky: As discussed here and on Slack, I don't agree that `noNullElements` is redundant, because it validates a different aspect of the inputs than `notEmpty` does.

Indeed, both `noNullElements` and `notEmpty` check for `options.params` _itself_ being `null` but that's not the point here. The point is that _only_ `noNullElements` explicitly checks whether any of the collection's _elements_ is `null` - as explained above, and on Slack, and as communicated by the method's names themselves.

As it is, when `options.params` has any of its elements `null`, `notEmpty` will pass (correctly), but the fact that there is a `null` element will be ignored without any clear message, and the validation will 'fall through' to the `allElementesOfType`. Luckily, this one will fail but with a message that will mislead into thinking that the failure was because not all elements were boolean, unnecessarily making any potential bug hunting a little harder.

In my view leaving the `noNullElements` validation off is unjustified because keeping it:
a) costs nothing (single, easy to read line of code - which was already there and is now actively being removed),
b) provides a chance to communicate what exactly has failed and why, through an explicit and accurate message that saves the reader time and effort of analysing code in order to find the cause,
c) does not incur any effort of writing test worth mentioning, because adding a test case to verify this validation is trivially simple.

Having said that, this is Monika's explicit decision made after hearing my arguments, therefore I accept this implementation and am resolving this discussion.

cms/src/test/java/uk/nhs/digital/apispecs/IsAnyTrueHelperTest.java
```diff
@@ -107,21 +109,39 @@ public void throwExceptionWhenCollectionIsNull() throws IOException {
     }
 
     @Test
-    public void throwExceptionWhenCollectionIsNotBoolean() throws IOException {
+    public void throwExceptionWhenAtLeastOneIsNull() throws IOException {
```
ziemsky, line 112: As mentioned in earlier comments and discussed on Slack, this test should fail when the method under test (`isAnyTrueHelper.apply`) fails to throw any exceptions (and the one we're asserting against in particular).

cms/src/test/java/uk/nhs/digital/apispecs/IsAnyTrueHelperTest.java
```diff
@@ -107,21 +109,39 @@ public void throwExceptionWhenCollectionIsNull() throws IOException {
...
+            assertThat("Exception in execution",
+                e.getMessage(),
+                containsString("All parameters are required to be boolean but at least one provided value was not.")
+            );
+        }
+    }
 
-        Object[] params = {false,"test", 4};
+    @Test
+    public void throwExceptionWhenAtLeastOneIsNonBoolean() throws IOException {
```
ziemsky, line 131: As mentioned in earlier comments and discussed on Slack, this test should fail when the method under test (`isAnyTrueHelper.apply`) fails to throw any exceptions (and the one we're asserting against in particular).

ziemsky (changes requested):

cms/src/test/java/uk/nhs/digital/apispecs/IsAnyTrueHelperTest.java
monika10793: They are failing. They don't need to be failed explicitly.

cms/src/test/java/uk/nhs/digital/apispecs/IsAnyTrueHelperTest.java
monika10793: They are failing. They don't need to be failed explicitly.

ziemsky: They are passing if the code of the method under test is modified so that it doesn't throw _any_ exceptions (not just the validations ones).

If you modify the method under test by embracing its _entire_ logic (including the validations) within this try-catch block, you'll see that this test and the `throwExceptionWhenAtLeastOneIsNonBoolean` one will pass.
```
try {
    ...all logic that could throw an exception is placed here...
} catch (Exception e) {
    e.printStackTrace();
}
```

Unlikely as it seems, this example is just one way to make the method not throw _any_ exceptions, but it's a common mistake made by developers who don't have an understanding of how exception handling should work.

The simple call to `fail()` method (like the one in the `throwExceptionWhenCollectionIsEmpty`) protects the code from this kinds of mistakes from occurring.

cms/src/test/java/uk/nhs/digital/apispecs/IsAnyTrueHelperTest.java
ziemsky: See my comment to `throwExceptionWhenAtLeastOneIsNull`.

monika10793: Fixed

cms/src/test/java/uk/nhs/digital/apispecs/IsAnyTrueHelperTest.java
monika10793: Fixed

ziemsky (approved):

PR #1084, opened 2020-11-09
DW-1583

LEJA3:
DW-1583: Fixed navigation arrow misalignment
Fixed wrapping inconsistency on hover

### DIFF
--- a/repository-data/webfiles/src/main/resources/site/src/scss/components/_navigation-tiles.scss
+++ b/repository-data/webfiles/src/main/resources/site/src/scss/components/_navigation-tiles.scss
@@ -111,24 +111,32 @@
 
 .navigation-tile__link-text {
     border-bottom: 1px solid;
+}
+.navigation-tile:hover .navigation-tile__link-text {
+    border-bottom-color: transparent;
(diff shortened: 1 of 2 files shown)


repository-data/webfiles/src/main/resources/site/src/scss/components/_navigation-tiles.scss
```diff
@@ -107,20 +107,20 @@
     line-height: 1.35;
     font-weight: bold;
     color: $blue;
+    margin-right: -4px;
+    margin-left: -4px;
 }
 
 .navigation-tile__link-text {
-    border-bottom: 1px solid;
+    text-decoration: underline;
```
GeoffHayward, line 115: Just a note, there could be an accessibility reason for a border being used over an underline.

GeoffHayward (approved):
Please double check the old method was not to do with accessibility. Other than that it looks okay to merge.

GeoffHayward (commented):
Just need squashing into one commit.

GeoffHayward (approved):

PR #1087, opened 2020-11-13
[ADZ-696] OAS schema anyOf field

ziemsky:
**Only commit `c76436e` needs reviewing.**

Commit `e07fa36` has already been reviewed and approved as part of ADZ-595 ticket/branch.

The reason there are two commits here is because the current branch was created off of another (ADZ-595) which has not been merged yet, and is currently under testing.

For the same reason the current branch has not been rebased onto `master` yet. First ADZ-595 will be merged and only then the current branch will be rebased.


cms/src/test/resources/test-data/api-specifications/SchemaHelperTest/schemaObjectCompleteTopLevel-simpleFields.json
```diff
@@ -0,0 +1,40 @@
...
+                                "deprecated": true,
+                                "description": "Test Schema Object description in `Markdown`.",
+                                "multipleOf": 1.2,
+                                "minimum": -2.3,
+                                "exclusiveMinimum": true,
+                                "maximum": 2.2,
+                                "exclusiveMaximum": false,
+                                "minLength": 11,
+                                "maxLength": 22,
+                                "pattern": "^(?:[0-9]{4}-[0-9]{2}-[0-9]{2})?(?:[ T][0-9]{2}:[0-9]{2}:[0-9]{2})?(?:[.,][0-9]{3})?",
```
GeoffHayward, line 22: Can a comment indicate what data type this is looking for?  I assume it's a date that can be empty.

cms/src/test/resources/test-data/api-specifications/SchemaHelperTest/schemaObjectsMultiLevelHierarchy-items.json
```diff
@@ -0,0 +1,110 @@
...
+                                            "deprecated": true,
+                                            "description": "Test Schema Object description in `Markdown`.",
+                                            "multipleOf": 1.2,
+                                            "minimum": -2.3,
+                                            "exclusiveMinimum": false,
+                                            "maximum": 2.4,
+                                            "exclusiveMaximum": true,
+                                            "minLength": 11,
+                                            "maxLength": 22,
+                                            "pattern": "^(?:[2])?",
```
GeoffHayward, line 30: Is the non-capture group needed? Wouldn't `^2?` do the same job?

GeoffHayward (approved):
Look great.

cms/src/test/resources/test-data/api-specifications/SchemaHelperTest/schemaObjectCompleteTopLevel-simpleFields.json
ziemsky: The short and cheeky answer is: I don't know and I don't care :grin: 

The reason is that, in context of this test (and in the rest of our implementation in the CMS), we don't use this value as a regex to do any matching; instead, all we do with it is just to display it on the website as is.

[OAS specification](https://spec.openapis.org/oas/v3.0.3#schema-object) requires that the string is a regex - and so I used a rather elaborate expression here but only to see that it is included in the HTML rendered by `SchemaHelper` class.

cms/src/test/resources/test-data/api-specifications/SchemaHelperTest/schemaObjectsMultiLevelHierarchy-items.json
ziemsky: The same situation as described in the previous comment - the actual value of the regex does not matter because all we do is display it rather than use it to match anything.

ziemsky:
Commit `810db3e` is just an update of this branch with the latest changes from `master`; functionally it's no different from the reviewed and approved `c76436e`.

ziemsky:
Commit `9a9443a` is just an update of this branch with the latest changes from `master`; functionally it's no different from the reviewed and approved `c76436e`.

PR #1092, opened 2020-11-19
DW 1616 move news year wise

Deven-Rawat:
Code added to move news from News And Event & News fodler to year of publication year.
Code also creates rules for old and new URL.
Also it creates a  csv file under asset folder


GeoffHayward (approved):
As long as it's been tested, it looks good to me.

Deven-Rawat:
This is done as part of DW-1617

PR #1094, opened 2020-11-23
[DW-1495] Image size handling

JodiWarren:


ShaneAhern-IRE (approved):
Approved.

Once deployed, please run the groovy scripts in the following order:

1. DW-1495-prepare-image-set-updater.groovy 
2. DW-1495-image-set-updater.groovy

PR #1095, opened 2020-11-24
[ADZ-700] OAS schema required

ziemsky:
Only `92edab1` needs reviewing; `17c7d5e` was reviewed as part of PR #1091 and `d59fa6a` was superseded.

PR #1091, opened 2020-11-18
[ADZ-699] oas schema field default

ziemsky:

PR #1098, opened 2020-11-25
[ADZ-599] OAS schema field 'enum'

ziemsky:
**Note**: only commit `90a9d7e` needs reviewing.

The earlier commits had already been reviewed as part of other pull requests and are only included here because the current branch was created from another which is undergoing tests and is yet to be merged back to `master`:
* `17c7d5e` reviewed as part of #1091,
* `92edab1` reviewed as part of #1095.

### DIFF
--- a/cms/src/main/java/uk/nhs/digital/apispecs/handlebars/EnumHelper.java
+++ b/cms/src/main/java/uk/nhs/digital/apispecs/handlebars/EnumHelper.java
@@ -9,11 +9,17 @@ import java.util.Collection;
 import java.util.Optional;
 import java.util.stream.Collectors;
 
-public class EnumHelper implements Helper<Collection<String>> {
+public class EnumHelper implements Helper<Collection<?>> {
 
-    public static final String NAME = "enum";
+    public static final EnumHelper INSTANCE = new EnumHelper();
 
-    @Override public Object apply(final Collection<String> args, final Options options) {
+    public static final String NAME = "enumeration";
+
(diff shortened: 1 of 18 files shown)


cms/src/main/java/uk/nhs/digital/apispecs/handlebars/EnumHelper.java
```diff
@@ -9,11 +9,17 @@
...
 import java.util.stream.Collectors;
 
-public class EnumHelper implements Helper<Collection<String>> {
+public class EnumHelper implements Helper<Collection<?>> {
 
-    public static final String NAME = "enum";
+    public static final EnumHelper INSTANCE = new EnumHelper();
 
-    @Override public Object apply(final Collection<String> args, final Options options) {
+    public static final String NAME = "enumeration";
```
ziemsky, line 16: The helper's name was changed from `enum` to `enumeration` to eliminate confusion: the relevant field in JSON is called `enum` and it could look puzzling to see the helper invoked in the template as `{{{enum enum}}}`.

GeoffHayward: Good idea.

GeoffHayward (approved):
Looks good.

ziemsky:
Commit `b57b407` contains no functional changes compared to the reviewed and approved `90a9d7e`. The forced push is just an update with the latest changes from `master`.

PR #1100, opened 2020-12-01
[DW-1635] Add Feed List component

JodiWarren:
Create Feed List component
Add Large Horizontal template
Add Small Horizontal template
Add Tiles template
Add optional title
Add optional button
Copy filter behaviour from sibling components
Ignore duplicated code


GeoffHayward (approved):
Looks good.

PR #1111, opened 2020-12-10
[DW-1836] Fixes sticky navigation

GeoffHayward:
Also:
Removes duplicate IDs (bug)
Removes unused code and imports


repository-data/webfiles/src/main/resources/site/src/js/sticky-nav/sticky-nav.js
```diff
@@ -1,41 +1,143 @@
...
+ * or out of view - then marks/unmasks the corresponding sticky nav list items
+ * appropriately.
+ *
+ * See above for details on 'logically nested' elements.
+ */
+export function initStickyNavSubItems() {
+    const io = new IntersectionObserver(entries => {
+        const getPreviousRelevantSibling = (elem, selector) => {
+            let sibling = elem.previousElementSibling
+            while (sibling) {
```
LEJA3, line 121: What's the function of `while` here? `previousElementSibling` should contain a single node or null.

LEJA3 (approved):

PR #1113, opened 2020-12-10
[ADZ-460] Try this API make optional

ziemsky:


ziemsky:
The only difference between `171591b` that the request was raised for, and the force-pushed `479f3a5` is that the latter adds declaration of the `handlebars-helpers` dependency to the `dependencyManagement` in the parent `pom.xml` and updates the branch with latest `master` changes.

GeoffHayward (approved):
Good work.

ziemsky:
Functionally, `be0e5b5` does not differ from `896a870`; the only reason for the forced push was to rebase the branch onto the latest `master`.

PR #1127, opened 2021-01-11
[DW-1790] Added time stamp to publication date time in cyber alert template

tomjdawson:
- UI changes to allow a timestamp to be added to the publish date within cyber alerts doc
- UI display of timestamp on website only for cyber alerts documents


GeoffHayward (approved):
Looks good

PR #1141, opened 2021-01-21
[DW-1544] Content search feature complete.

ShaneAhern-IRE:


conf/content-search-cms.properties
```diff
@@ -0,0 +1 @@
+brx.contentfeed.environment = dev-localhost
```
ShaneAhern-IRE, line 1: Remove - only needed on servers

GeoffHayward (approved):
Super work.

PR #1151, opened 2021-01-27
[DW-1529] Add self-service deployments to non-prod

GeoffHayward:


ziemsky (approved):
Nice job and a cool feature :+1:

PR #1156, opened 2021-01-29
Adz 483

harryhmc:
PR covering changes for ADZ-483.

To get around the issue of changes to rendering logic not triggering the re-rendering of pages, a new scheduled job has been introduced that converts the JSON in a document, compares it to what is published, and publishes the new version if necessary.

### DIFF
--- a/cms/src/main/java/uk/nhs/digital/apispecs/module/ApiSpecSyncFromApigeeModule.java
+++ b/cms/src/main/java/uk/nhs/digital/apispecs/module/ApiSpecSyncFromApigeeModule.java
@@ -1,19 +1,19 @@
 package uk.nhs.digital.apispecs.module;
 
-import static uk.nhs.digital.apispecs.module.ApiSpecSyncFromApigeeModule.ConfigPropertyNames.CRON_EXPRESSION;
-import static uk.nhs.digital.apispecs.module.ApiSpecSyncFromApigeeModule.ConfigPropertyNames.SYNC_ENABLED;
-
 import org.onehippo.cms7.services.HippoServiceRegistry;
 import org.onehippo.repository.modules.AbstractReconfigurableDaemonModule;
 import org.onehippo.repository.modules.RequiresService;
+import org.onehippo.repository.scheduling.RepositoryJob;
 import org.onehippo.repository.scheduling.RepositoryJobCronTrigger;
 import org.onehippo.repository.scheduling.RepositoryJobInfo;
 import org.onehippo.repository.scheduling.RepositoryScheduler;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.nhs.digital.JcrNodeUtils;
+import uk.nhs.digital.apispecs.jobs.ApiSpecRerenderJob;
 import uk.nhs.digital.apispecs.jobs.ApiSpecSyncFromApigeeJob;
 
+import java.util.Arrays;
 import java.util.Optional;
 import java.util.stream.Stream;
 import javax.jcr.Node;
@@ -28,7 +28,6 @@ public class ApiSpecSyncFromApigeeModule extends AbstractReconfigurableDaemonMod
     private final Object configurationLock = new Object();
 
     private static final String JOB_GROUP = "devzone";
-    private static final String JOB_NAME = "apiSpecSyncFromApigee";
     private static final String JOB_TRIGGER_NAME = "cronTrigger";
 
     @Override protected void doConfigure(final Node moduleConfigNode) {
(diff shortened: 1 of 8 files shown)


cms/src/main/java/uk/nhs/digital/apispecs/module/ApiSpecSyncFromApigeeModule.java
```diff
@@ -39,22 +61,23 @@
         // job here rather than in doInitialize which is only being called on system startup.
 
         synchronized (configurationLock) {
+            for (final String jobName: JOB_NAMES) {
+                try {
+                    log.info("Configuring job {}/{} for synchronising API Specifications from Apigee.", JOB_GROUP, jobName); // the only one called on the config node change
```
ziemsky, line 66: Log content probably wants to be more generic now that we have added a job that is _not_ about synchronising from Apigee. How about something like?:

> Configuring job {}/{} for updating API specifications.

cms/src/main/java/uk/nhs/digital/apispecs/module/ApiSpecSyncFromApigeeModule.java
```diff
@@ -28,8 +32,26 @@
     private final Object configurationLock = new Object();
 
     private static final String JOB_GROUP = "devzone";
-    private static final String JOB_NAME = "apiSpecSyncFromApigee";
     private static final String JOB_TRIGGER_NAME = "cronTrigger";
+    private static final String DAILY_JOB_NAME = "apiSpecSyncFromApigee";
+    private static final String NIGHTLY_JOB_NAME = "apiSpecRefresh";
+    private static final String[] JOB_NAMES = {DAILY_JOB_NAME, NIGHTLY_JOB_NAME};
```
ziemsky, line 38: I see the purpose behind `JOB_NAMES`, `CRON_INDEX` and `CLASS_INDEX` but, I think, we could make it a bit simpler.

Each of these structures (including the `ConfigPropertyNames` enum) contains a collection of job-specific properties, but, at the moment, these are scattered across those disparate collections. By aggregating these properties in one 'collection' keyed on a job type we could make this a bit more object-oriented and easier to follow.

What I'd do is re-define the enum as something like:
```java
    enum JOBS {

        DAILY(
            "apiSpecSyncFromApigee",
            ApiSpecSyncFromApigeeJob.class,
            "devzone.apispec.sync.daily-cron-expression",
            "cronExpression"
        ),
        NIGHTLY(
            "apiSpecRefresh",
            ApiSpecRerenderJob.class,
            "devzone.apispec.sync.nightly-cron-expression",
            "cronExpression"
        );

        private final String systemPropertyName;
        private final String jcrPropertyName;
        private final String jobName;
        private final Class<?> jobClass;

        JOBS(final String jobName, final Class<?> jobClass, final String systemPropertyName, final String jcrPropertyName) {
            this.systemPropertyName = systemPropertyName;
            this.jcrPropertyName = jcrPropertyName;
            this.jobName = jobName;
            this.jobClass = jobClass;
        }
    }
```
...where `doConfigure` could then be turned into something like this:
```java
@Override protected void doConfigure(final Node moduleConfigNode) {

        // ...

        synchronized (configurationLock) {
            Arrays.stream(JOBS.values()).forEach(job -> {
                try {
                    log.info("Configuring job {}/{} for updating API Specifications from Apigee.", JOB_GROUP, job.jobName); // the only one called on the config node change

                    deactivatePreviousJobInstanceIfExist(job);

                    if (jobShouldBeEnabled(moduleConfigNode, job)) {

                        scheduleNewJob(moduleConfigNode, job);

                    } else {
                        log.warn("Job {}/{} is not configured to be enabled, therefore it will not be activated.", JOB_GROUP, job.jobName);
                    }

                } catch (final Exception e) {
                    throw new RuntimeException(String.format("Failed to configure job %s/%s", JOB_GROUP, job.jobName), e);
                }
            });
        }
    }
```
...updating the methods called from it to accept single `job` enum value and retrieve job's properties from it.

harryhmc: love this - i wasn't happy with the use of two Maps. hadn't considered expanding the enum but makes perfect sense :+1:

ziemsky (commented):
Nice work!

cms/src/main/java/uk/nhs/digital/apispecs/module/ApiSpecSyncFromApigeeModule.java
```diff
@@ -64,61 +64,58 @@
...
-        final Optional<Boolean> enabledViaSystemProperty =
-            Optional.ofNullable(System.getProperty(SYNC_ENABLED.systemPropertyName)).map(Boolean::parseBoolean);
+        final Optional<String> cronExprFromSystemProperty = Optional.ofNullable(System.getProperty(job.systemPropertyName));
 
-        return Stream.of(enabledViaJcr, enabledViaSystemProperty)
-            .filter(Optional::isPresent)
-            .map(Optional::get)
-            .findFirst()
-            .orElse(false);
+        return Stream.of(cronExprFromJcr, cronExprFromSystemProperty).anyMatch(Optional::isPresent);
```
ziemsky, line 111: Ooh, nice! Much cleaner than the old version. I'm stealing that!

vars.mk
```diff
@@ -22,8 +23,14 @@ S3_REGION ?= eu-west-1
...
-APIGEE_SYNC_ENABLED ?= false
-APIGEE_SYNC_CRON ?= 0 0/1 * ? * *
+# Both cron expression variables below will need to have a property set in the system to
+# enable the scheduled jobs in a local CMS instance. In order to keep this out of version
+# control, we recommend setting both variables in env.mk as needed, e.g.
+# 	APIGEE_SYNC_CRON = 0 0/2 * ? * *
+# 	APIGEE_RERENDER_CRON = 0 0/10 * ? * *
+
+APIGEE_SYNC_CRON ?=
+APIGEE_RERENDER_CRON ?=
```
ziemsky, line 33: :+1:

cms/src/test/java/uk/nhs/digital/apispecs/ApiSpecificationPublicationServiceTest.java
```diff
@@ -188,6 +188,59 @@ public void publish_canHandleMultipleSpecifications() {
...
+
+        // then
+        then(cmsSpecPublished).should().getSpecJson();
+        then(cmsSpecPublished).should().getHtml();
+        then(cmsSpecPublished).should().getId(); // this call is made for the sake of logging
+        then(cmsSpecPublished).shouldHaveNoMoreInteractions();
+    }
+
+    @Test
+    public void refresh_updatesAndPublishesSpecificationWhenChanged() {
```
ziemsky, line 220: `refresh` wants to be `rerender`?

cms/src/test/java/uk/nhs/digital/apispecs/module/ApiSpecSyncFromApigeeModuleTest.java
```diff
@@ -23,16 +24,20 @@
...
 import javax.jcr.Node;
 import javax.jcr.RepositoryException;
 
 @RunWith(PowerMockRunner.class)
 @PrepareForTest({HippoServiceRegistry.class, ApiSpecSyncFromApigeeModule.class})
 public class ApiSpecSyncFromApigeeModuleTest {
 
-    public static final String JOB_NAME = "apiSpecSyncFromApigee";
+    public static final String DAILY_JOB_NAME = "apiSpecSyncFromApigee";
+    public static final String NIGHTLY_JOB_NAME = "apiSpecRefresh";
```
ziemsky, line 40: `apiSpecRefresh` wants to be `apiSpecRerender` as it breaks the test, otherwise.

ziemsky (approved):
It seems that asking to replace the remaining 'refresh' words with 'rerender' I've made you break a test. Sorry about that :grin:

PR #1159, opened 2021-02-01
[ADZ-343] Balance specification headings

ziemsky:


cms/src/main/java/uk/nhs/digital/apispecs/handlebars/HeadingsHyperlinksFromMarkdownHelper.java
```diff
@@ -0,0 +1,245 @@
...
+
+    private HeadingsHyperlinksFromMarkdownHelper() {
+        // no-op; made private to promote the use of INSTANCE
+    }
+
+    @Override
+    public Options.Buffer apply(final String markdown, final Options options) {
+
+        try {
+            return hyperlinksRenderedFromHeadingsIn(markdown, options);
```
harryhmc, line 93: maybe just personal preference on my end, but IMO single-use single-line functions are harder to follow than inline logic and dont provide valuable abstraction / maintainability / test-ability benefits. there's also a lot of nested function calls in this file. unless you feel strongly i would prefer this to be written as
```
List<HeadingModel> headingModels = headingModelsFrom(markdown);
return render(headingModels, options);
```

harryhmc (approved):

cms/src/main/java/uk/nhs/digital/apispecs/handlebars/HeadingsHyperlinksFromMarkdownHelper.java
ziemsky: Thanks! I totally get the point.

I often use this convention in 'framework' methods with generic names that I override and therefore cannot change, as a way to express what exactly is the intent of the logic invoked by said generic method.

It could easily be argued that it was an overkill here and I'm not precious about it. Fixed now in the overriding, force-pushed commit.

cms/src/main/java/uk/nhs/digital/apispecs/commonmark/CommonmarkMarkdownConverter.java
```diff
@@ -65,40 +75,162 @@ public String toHtml(final String markdown) {
...
+     *     ### Heading B
+     *     ###### Heading C
+     * </pre>
+     * ...will be rendered as:
+     * <pre>
+     *     &lt;h3 id=&quot;heading-a&quot;&gt;Heading A&lt;/h3&gt;
+     *     &lt;h5 id=&quot;heading-b&quot;&gt;Heading B&lt;/h5&gt;
+     *     &lt;h8 id=&quot;heading-c&quot;&gt;Heading C&lt;/h8&gt;
+     * </pre>
+     * <p>
```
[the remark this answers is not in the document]
ziemsky: Well spotted!

That was deliberate to reduce the clutter a little bit - IntelliJ appears to be rendering the JavaDoc fine without them, and this is the only context those comments are ever being rendered as actual HTML so I'm inclined to leave these as they are if it's all the same to you.

JavaDoc rendering in IntelliJ

PR #1167, opened 2021-02-04
[DW-000] Updating sitemap for cyber alerts detailed page

ShaneAhern-IRE:
Fix for cyber alerts detailed page.


GeoffHayward (approved):
Many thanks.

PR #1170, opened 2021-02-09
[ADZ-627] Added API catalogue template and sitemap configuration

harryhmc:


acceptance-tests/src/test/java/uk/nhs/digital/ps/test/acceptance/AcceptanceTest.java
```diff
@@ -9,7 +9,7 @@
  */
 @RunWith(Cucumber.class)
 @CucumberOptions(
-        features = "src/test/resources/features",
+        features = "src/test/resources/features/site",
```
ziemsky, line 12: I guess the `/site` suffix was committed unintentionally?

acceptance-tests/src/test/java/uk/nhs/digital/ps/test/acceptance/steps/site/SiteSteps.java
```diff
@@ -199,6 +200,38 @@ public void thenIShouldSeeTheFollowing(DataTable contentsTable) throws Throwable
...
+            String xPathExpression = buildXPathExpression(keys, elementItem);
+
+            assertNotNull(
+                "I should find element with xpath expression " + xPathExpression,
+                sitePage.findElementWithXPath(xPathExpression)
+            );
+        }
+    }
+
+    private String buildXPathExpression(List<String> keys, List<String> elementItem) {
```
ziemsky, line 219: The method name could probably be more specific, reflecting some details of the expression that it builds.

Not that I can think of any good name myself, mind you :grin:

acceptance-tests/src/test/resources/features/site/apiCatalogue.feature
```diff
@@ -0,0 +1,26 @@
+Feature: API Catalogue in Developer hub
+    @WIP
```
ziemsky, line 2: `@WIP` can now be removed.

acceptance-tests/src/test/resources/features/site/apiCatalogue.feature
```diff
@@ -0,0 +1,26 @@
...
+    Scenario: Catalogue is rendered in Site
+        Given I navigate to "Static API Catalogue" page
+        Then I should see page titled "Test API catalogue"
+        And I can see labelled element "website.linkslist.summary" with content "These are the APIs which are currently available, or in development, from NHS Digital."
+        And I should see the list with title "Alphabetical navigation" containing:
+            | A | B | C | D | E | F | G | H | I | J | K | L | M | N | O | P | Q | R | S | T | U | V | W | X | Y | Z |
+        # Assert that index is rendered correctly
+        And I can see elements with attributes:
```
ziemsky, line 10: I wonder if we could add an overloaded method here with name and text in `@Then` annotation reflecting the text in the comment. It would sit alongside and invoke the `I can see elements with attributes` method.

We could then get rid of the comment line above and have the method express the intent itself.

acceptance-tests/src/test/resources/features/site/apiCatalogue.feature
```diff
@@ -0,0 +1,26 @@
...
+            | text | href  | aria-label                                    |
+            | B    | #b    | Jump to articles starting with the letter 'B' |
+            | G    | #g    | Jump to articles starting with the letter 'G' |
+            | N    | #n    | Jump to articles starting with the letter 'N' |
+        # Assert that links are rendered correctly
```
ziemsky, line 15: Same as above.

site/components/src/main/java/uk/nhs/digital/common/components/ApiCatalogueComponent.java
```diff
@@ -0,0 +1,18 @@
...
+
+import java.util.Optional;
+
+public class ApiCatalogueComponent extends BaseGaContentComponent {
+
+    @Override public void doBeforeRender(final HstRequest request,
+                                         final HstResponse response) {
+        super.doBeforeRender(request, response);
+
+        Optional.ofNullable(request.getRequestContext().getContentBean()).ifPresent(document -> {
```
ziemsky, line 14: I'd be tempted to try removing this entire statement and see what happens.

One of the methods in one of the the super classes already invokes `request.setModel` which, I think, achieves the same objective.

Either way, we can keep the current class itself because we'll soon need it when we add support for filtering by taxonomy.

site/components/src/main/java/uk/nhs/digital/website/beans/ApiCatalogue.java
```diff
@@ -0,0 +1,17 @@
+package uk.nhs.digital.website.beans;
+
+import org.hippoecm.hst.content.beans.Node;
+import org.onehippo.cms7.essentials.dashboard.annotations.HippoEssentialsGenerated;
+
+@Node(jcrType = "website:apicatalogue")
+@HippoEssentialsGenerated(internalName = "website:apicatalogue", allowModifications = false)
+public class ApiCatalogue extends CommonFieldsBean {
```
ziemsky, line 8: We definitely don't need the current getter methods because they have no meaning in context of API Catalogue document.

I know that this leaves the class empty but am not sure if we can remove it completely, or would that break the rendering?

ziemsky (changes requested):

acceptance-tests/src/test/java/uk/nhs/digital/ps/test/acceptance/steps/site/SiteSteps.java
harryhmc: `buildXPathExpressionFromElementAttributes`? `buildXPathExpressionForElement`?

site/components/src/main/java/uk/nhs/digital/website/beans/ApiCatalogue.java
harryhmc: i assume we need the class to be here (even if empty) since `dynamic_bean_generation` is set to false

once we have introduce the taxonomy filter, we'll probably need a method here to return the listcomponent of internal links so i'll just leave it for now?

acceptance-tests/src/test/java/uk/nhs/digital/ps/test/acceptance/steps/site/SiteSteps.java
ziemsky: `buildXPathExpressionFromElementAttributes` probably sounds better of the two :+1:

site/components/src/main/java/uk/nhs/digital/website/beans/ApiCatalogue.java
ziemsky: Indeed, good point. No harm leaving it in. At the very least, it gives the bean a more precise name.

ziemsky (dismissed):

ziemsky (approved):

PR #1177, opened 2021-02-17
[DW-1507] Search badges correctly show up for blogs

robertSahakyan:


iainstarks (commented):
add an enum value for blog and use it on a Blog type - seems pretty obvious

GeoffHayward (approved):

PR #1180, opened 2021-02-19
add includeSubDomains for security header Strict-Transport-Security f…

iainstarks:
as per the ticket, the security headers were amost as required with the exception that there is an includeSubDomains property on the content-security-policy header


iainstarks:
Agreed not to merge this as the includeSubDomains propery on strict-transport-security is not needed

PR #1186, opened 2021-02-24
[DW-2049] Files added for UI toolkit

mehranrashid:
Changes made so the UI toolkit can be utilised in the intranet re-branding


iainstarks (commented):
this is to include the new toolkit which has been built and copied from frontend,  so that future intranet tickets can use it - we've sanity checked intranet and main site with this on dev

repository-data/webfiles/src/main/resources/site/src/scss/nhsd-intranet.scss
```diff
@@ -1,10 +1,12 @@
 // The Intranet styling system based on the NHSD styling system!
 @import "nhsuk";
```
GeoffHayward, line 2: I see you are still importing the main website's CSS. Long-term these should be 100% isolated projects (i.e. CSS/JS for website and separate CSS/JS for intranet).  Then the common CSS/JS, used by both the website and the interment, comes in from the NHSD UI Toolkit.

GeoffHayward (commented):
I have a question about 3 of the files. The following are minified. Should they be Git tracked?
- intranet.js
- intranet-components.scss 
- intranet.scss

GeoffHayward (approved):

PR #1229, opened 2021-03-22
DW-2206 bespoke rewriter

mehranrashid:
A rewriter class that styles the p tags for the new brand refresh


GeoffHayward (approved):
Looks good to me. Nice work.

PR #1231, opened 2021-03-23
Dw 1624 move events feed

lekeDevOps:
Hi I think @Deven-Rawat you might be best placed, as @naimullakhan1 mentioned you were working on something similar....

Would you do the honours of reviewing this?

KR
Lesley


lekeDevOps:
this has been merged into 1617, therefore can be deleted.

PR #1268, opened 2021-04-12
[ADZ-992] Make rendering not require 'components' field

urchy:


cms/src/main/java/uk/nhs/digital/apispecs/ApiSpecificationPublicationService.java
```diff
@@ -21,13 +21,14 @@
...
 public class ApiSpecificationPublicationService {
 
     private static final Logger log = LoggerFactory.getLogger(ApiSpecificationPublicationService.class);
 
-    private OpenApiSpecificationJsonToHtmlConverter openApiSpecificationJsonToHtmlConverter;
-    private OpenApiSpecificationRepository remoteSpecRepository;
-    private ApiSpecificationDocumentRepository localSpecRepository;
+    private final OpenApiSpecificationJsonToHtmlConverter openApiSpecificationJsonToHtmlConverter;
+    private final OpenApiSpecificationRepository remoteSpecRepository;
+    private final ApiSpecificationDocumentRepository localSpecRepository;
```
ziemsky, line 31: :+1:

cms/src/test/java/uk/nhs/digital/apispecs/SwaggerCodeGenOpenApiSpecJsonToHtmlConverterTest.java
```diff
@@ -62,6 +62,25 @@ public void getHtmlForSpec_rendersCompleteOpenApiJsonAsHtml() {
         );
     }
 
+    @Test
+    public void getHtmlForSpec_rendersIncompleteOpenApiJsonAsHtml() {
```
ziemsky, line 66: This test method targets a very specific aspect of the specification source so it would be good to express that in the methods' name. Simply saying 'incomplete' feels a bit too general and one has to look at the JSON file name to figure out that we specifically care about the `components` field missing.

How about naming it along the lines of `rendersSpecification_withoutFailure_whenComponentsPropertyIsAbsent`?

cms/src/test/java/uk/nhs/digital/apispecs/SwaggerCodeGenOpenApiSpecJsonToHtmlConverterTest.java
```diff
@@ -62,6 +62,25 @@ public void getHtmlForSpec_rendersCompleteOpenApiJsonAsHtml() {
...
+        final String incompleteSpecificationJson = from("oasV3_incomplete_no_components-field.json");
+
+        final String expectedSpecHtml = from("oasV3_complete.html");
+
+        // when
+        final String actualSpecHtml = swaggerCodeGenApiSpecHtmlProvider.htmlFrom(incompleteSpecificationJson);
+
+        // then
+        assertThat(
+            "A complete spec has been rendered as HTML using an incomplete specification customised Swagger CodeGen v3",
```
ziemsky, line 78: Similar comment to that about the method name: the assertion message could probably be more accurate if it said something like `Specification is rendered with no error when the 'components' field is absent from the source JSON`.

cms/src/test/java/uk/nhs/digital/apispecs/SwaggerCodeGenOpenApiSpecJsonToHtmlConverterTest.java
ziemsky: Since we're changing the method name: nothing to do with the current story but I've just noticed that a number of method names in this class start with `getHtmlForSpec` - it appears that, at some point, we may have changed the name of the method these tests invoke but have forgotten to update names of methods in this class.

Would you mind updating the names of the remaining methods here? I think we can drop the `getHtmlForSpec` part altogether and stick to the pattern of `doesSomething_whenSomethingElse`.

urchy: Thanks Rafal, you right.  The name of testing must be specific to the case that we are testing. I'll change it.

ziemsky (changes requested):
Nice one!

Good to see that the issue that we're fixing here is covered with a dedicated test that demonstrates that it's, indeed, been fixed :+1: 

A couple of cosmetic changes suggested.

ziemsky (approved):
:+1:

PR #1280, opened 2021-04-20
[ADZ-903] Multiple filters in API catalogue

ziemsky:
It's now possible to apply multiple filters in the API catalogue
at the same time.

Also:
* Makes the rendering of API catalogue page not fail
  when parsing of taxonomy-to-filters mapping fails;
  Filters section will be missing but catalogue's
  content will still display, in such a case.
* Fixes a bug in TestLoggerRule, where it was swallowing
  certain errors in tests.
* Moves logging test utility classes to the shared
  `test-resources` module so that they can be used both
  from CMS as well as Site modules.
* Upgrades ChromeDriver version.


cms/src/test/java/uk/nhs/digital/apispecs/ApiSpecificationPublicationServiceTest.java
```diff
@@ -325,9 +325,9 @@ public void sync_handlesMultipleSpecifications() {
 
         // published before but not changed in Apigee - expect to not be updated nor published
         final String publishedNotChanged_id                = "publishedNotChanged";
-        final String publishedNotChangedRemote_modTime     = "2020-05-10T10:30:00.001Z";
-        final String publishedNotChangedOld_checkTime      = "2020-05-10T10:30:00.002Z";
-        final Instant publishedNotChanged_newCheckTime     = Instant.parse("2020-05-10T10:30:00.002Z");
```
harryhmc, line 330: not sure why these changes have been made?

repository-data/webfiles/src/main/resources/site/freemarker/common/api-catalogue.ftl
```diff
@@ -49,23 +50,38 @@
...
                     <#if filtersModel?? && !filtersModel.isEmpty()>
                         <div class="article-section-nav-wrapper">
-                            <div class="article-section-nav">
-                                <h2 class="article-section-nav__title">Filters</h2>
-                                <nav class="filters">
+                            <div class="article-section-nav filters">
+
+                                <h2 class="article-section-nav__title">
+                                    <span class="filter-head__title">Filters</span>
+                                    <@hst.renderURL var="resetUrl"/>
```
harryhmc, line 62: could you explain what this macro does?

repository-data/webfiles/src/main/resources/site/freemarker/common/api-catalogue.ftl
```diff
@@ -49,23 +50,38 @@
...
                                             <li>
-                                                <input id="toggler_${section.key}" type="checkbox" <#if section.expanded>checked</#if>/>
-                                                <label for="toggler_${section.key}" class="section-heading <#if section.expanded>selected</#if>">${section.displayName}</label>
+                                                <input id="toggler_${section.key}" type="checkbox" <#if section.expanded>checked</#if> aria-hidden="true"/>
+                                                <#if section.expanded>
+                                                    <#assign "expansionArrowClass" = "expansion-arrow selected" />
+                                                <#else>
+                                                    <#assign "expansionArrowClass" = "expansion-arrow" />
+                                                </#if>
+                                                <@svgIcon "expantionArrow" expansionArrowClass/>
```
harryhmc, line 77: typo in 'expansion' here (i know it's consistent with where you've declared the svgIcon, but even so could this be fixed pls)

repository-data/webfiles/src/main/resources/site/freemarker/common/api-catalogue.ftl
```diff
@@ -49,23 +50,38 @@
...
+                                <nav>
                                     <ul>
                                         <#list filtersModel.sections as section>
                                             <#if section.displayed>
                                             <li>
-                                                <input id="toggler_${section.key}" type="checkbox" <#if section.expanded>checked</#if>/>
-                                                <label for="toggler_${section.key}" class="section-heading <#if section.expanded>selected</#if>">${section.displayName}</label>
+                                                <input id="toggler_${section.key}" type="checkbox" <#if section.expanded>checked</#if> aria-hidden="true"/>
+                                                <#if section.expanded>
+                                                    <#assign "expansionArrowClass" = "expansion-arrow selected" />
```
harryhmc, line 73: does the following work in place of this if/else block?
```
<#assign "expansionArrowClass" = "expansion-arrow <#if section.expanded>selected</#if>" />
```

repository-data/webfiles/src/main/resources/site/src/scss/components/_api-catalogue.scss
```diff
@@ -58,17 +86,23 @@
 
     & .section-heading:hover,
     & .filter-label:hover {
-      background: $yellow-link-hover;
+      background: $yellow-link-hover !important;
       box-shadow: 0 0 0 4px $yellow-link-hover;
       outline: 1px solid $yellow-link-hover;
-      color: $link-hover-colour-nhs;
+      color: $link-hover-colour-nhs !important;
```
harryhmc, line 92: are these being overwritten by the .section-heading.selected / .filter-label.selected blocks below? if so maybe they could be rewritten as
```
.section-heading.selected,
.filter-label.selected {
  font-weight: bold;
}

.section-heading.selected:not(:hover) {
  // set colour properties here
}
// same for filter-label
```

up to you whether you feel it's necessary or not, i'd just rather avoid using `!important` if possible :stuck_out_tongue_closed_eyes:

site/components/src/main/java/uk/nhs/digital/common/components/apicatalogue/filters/Filters.java
```diff
@@ -40,15 +56,27 @@
         return allSectionsFlattened;
     }
 
-    public Filters initialisedWith(
-        final Set<String> filteredTaxonomyTags,
-        final Set<String> selectedTags
-    ) {
-        final FilterVisitor visitor = new StatusUpdatingFilterVisitor(filteredTaxonomyTags, selectedTags);
+    // Invoked from the template, hence shown by IDE as not used (but still needed).
```
harryhmc, line 59: i like these comments :smile:

harryhmc (commented):
LGTM, just a few questions i'd like answered before i give it a thumbs up

cms/src/test/java/uk/nhs/digital/apispecs/ApiSpecificationPublicationServiceTest.java
ziemsky: The reason for shifting a few string date values to the right was to ensure they align with the values that are preceded by `Instant.parse`, so that it's easier to see the differences between them when all the components of the dates are in the same column (values that are close together only differ in single component). The test values in this method are difficult to grasp as they are, so this is my attempt to make it easier, if only by a smidgen.

The changes within the values themselves was to move the differences from milliseconds to seconds, to ensure that minuscule timing variations in tests don't produce varying results. Admittedly, this particular change does not appear to be needed. The reason I put it in place was because there is a weird difference in how IDE runs the tests vs Maven. This particular test fails in my IDE (but only when run in a larger group of tests) but passes in Maven; fixing it for IDE makes it fail in Maven. The too-small-differences in the time values were my initial suspected culprit so I increased them from milliseconds to seconds. At the end, I didn't resolve the problem (the test passes in Maven which is what we care about more than IDE) but left the changes in place.

Happy to revert the latter change, if you like?

repository-data/webfiles/src/main/resources/site/freemarker/common/api-catalogue.ftl
ziemsky: It saves the value of the URL of the current page in a template variable whose name is given via `var`. This variable can then be used to render the URL, say, as the `<a href=` value (which is what we do in the next line with the `resetUrl` variable).

The URL contains the domain and path of the page but does not contain any query parameters. Given that we pass selected filters as query parameters, opening the page through a URL with no parameters has the effect of 'unsetting' any filters that may have been set - which is what this implementation of the _Reset_ button does.

It's part of the 'standard' tag library included by Bloomreach; they are actually JSP tags but can be used from Freemarker. For reference:

- [Bloomreach page that mentions the tags](https://documentation.bloomreach.com/14/library/concepts/web-application/hst2-tag-library.htmlframework).
- [Full collection of available tags](https://javadoc.onehippo.org/13.0/tlddocs/)
- [renderUrl tag](https://javadoc.onehippo.org/13.0/tlddocs/hst-core-tags/renderURL.html)

repository-data/webfiles/src/main/resources/site/freemarker/common/api-catalogue.ftl
ziemsky: Indeed! Your version is much cleaner; thanks, Harry :+1: 

I'm changing it now.

ziemsky: As it turns out, strings can't be interpolated with the `<#if...>` constructs like this.

However, Freemarker contains a useful '[built-in|(https://freemarker.apache.org/docs/dgui_template_valueinsertion.html#autoid_17)', which, combined with string contactenation, works as a good equivalent of Java's ternary operator. The following resulting line seems to works:

`<#assign "expansionArrowClass" = "expansion-arrow " + section.expanded?string("selected", "") />`

I'm glad you gave me an excuse to look into it a bit more :wink:

repository-data/webfiles/src/main/resources/site/freemarker/common/api-catalogue.ftl
ziemsky: Good spot! Indeed, this would bother me too. I'm changing it now.

ziemsky:
`4cd7ec1` applies changes requested in the code review.

ziemsky:
`16db92b` fixes `SyncResults` - it was collecting sync stats into a `Set<SyncResult>`. This resulted in `uk.nhs.digital.apispecs.ApiSpecificationPublicationServiceTest#sync_handlesMultipleSpecifications` to flicker as the order of log statements was not constant because the order of elements in the set changed between runs (`SyncResult` does not implement its own hashCode method); changing the field type from `Set` to `List` eliminated this problem and ensured that the stats are reliably logged in the order they are collected in.

`eac5cbe` updates ChromeDriver's version.

cms/src/test/java/uk/nhs/digital/apispecs/ApiSpecificationPublicationServiceTest.java
harryhmc: sure, i don't have a problem with the latter change but if it's not fixing the problem behaviour you're seeing, may as well revert it :)

ziemsky: Good point; the changes to values have now been reverted.

ziemsky:
`81ec3a5` reverts redundant changes to the values of timestamps in `ApiSpecificationPublicationServiceTest#sync_handlesMultipleSpecifications` as per Harry's comment.

harryhmc (approved):

PR #1298, opened 2021-04-22
[ADZ- 861] -  API Spec make side nav handle long list

urchy:


ziemsky (approved):
The new `<div>` element (including its enclosing tag) could probably use re-aligning for the hierarchy of elements to be more consistent, but that's no reason to not approve, this is just me being too pedantic :grin:

ziemsky:
Latest changes look good - thanks for indulging my OCD with re-aligning those HTML elements :grin:

PR #1337, opened 2021-04-30
[ADZ-613] API Spec: Try this API: eliminate the need for the 'refresh spec page' disclaimer

urchy:


repository-data/webfiles/src/main/resources/site/freemarker/common/api-try-it-now.ftl
```diff
@@ -187,13 +188,9 @@
...
                 <#else>
-                    <#assign section = {
-                        "emphasisType": "Important",
-                        "heading": "Open from the API Specification page",
-                        "bodyCustom":
-                        "Please open this feature via 'Try this API' button in the page of the API Specification. If you did just that and still see this message, please refresh the specification page and try opening this popup again."
-                    } />
-                    <@emphasisBox section=section />
+                    <script>
+                        window.location.href = '/site/error/404';
```
ziemsky, line 192: As it, this will only work when running on a local machine and break when running on any of the servers; the reason is that on the servers, the `/site` prefix is not included.

The following _should_ (subject testing) make this work in all environments:
`window.location.href = '<@hst.link siteMapItemRefId='pagenotfound'/>';`

The `pagenotfound` points to `sitemap`'s [/error/404](https://github.com/NHS-digital-website/hippo/blob/18da9ebfdd753730a81f5f3d9b641c9b8712cb9a/repository-data/site/src/main/resources/hcm-config/hst/configurations/common/sitemap.yaml#L349), identified by its `hst:refId: pagenotfound` and the `<@hst.link` macro dynamically derives correct URL for error page.

urchy: Make sense. Thanks Rafal. 
I'll update it. 

I might need your help to test it in different environments.

PR #1346, opened 2021-05-04
[ADZ-1097] ADZ-1097 removed pointer-events from section arrows

harryhmc:
setting `pointer-events: none` in the CSS class for section arrows so that click events are passed down to the section label underneath

[as seen in the MDN web docs](https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events#browser_compatibility), the `pointer-events` attribute is supported by all major browsers in both web and mobile versions

unrelated to ADZ-1097, i have set a column width on the button_column in schema tables to preserve right alignment of toggler icons


ziemsky (approved):
:+1:

PR #1376, opened 2021-05-13
[ADZ-322] Automatic try-it-out trigger

urchy:
Override swagger-ui configuration allows us to hide the try-it-now configurations and enable all input fields for params.


ziemsky (approved):
Great work! Exactly what we needed :+1:

PR #1424, opened 2021-05-19
ADZ-439 path parameters propagation

ziemsky:
[ADZ-439] OAS path parameters propagation

ADZ-422 ensured that path parameters propagate
to operations where operations don't define
any of their own. It achieved that by overriding
a method of CodeGen's `DefaultGenerator`. The
latter is not designed for extensions, and
this required duplicating a lot of code in the
extending class to apply the fix.

It has now turned out, however, that the required
behaviour can be toggled through configuration,
and so this change applies the configuration
in place of the override (see the use of
`ParserConfiguration` in
`SwaggerCodeGenOpenApiSpecificationJsonToHtmlConverter`.

This change also:
* Upgrades Codegen components to their
  latest versions.
* Moves the change introduced by ADZ-992 from
  `CodegenDefaultGenerator` to
  `ApiSpecificationStaticHtml2Codegen`.
* Removes the now redundant `CodegenDefaultGenerator`
  class.
* Reduces delay with which API catalogue related
 jobs are scheduled down to 10 minutes, given
 that, according to Rundeck, the redeployment
 does not take more than that interval.
* Cleans up `SwaggerCodeGenOpenApiSpecificationJsonToHtmlConverter`
 making the flow more readable.


urchy (approved):
One of our most important classes is so much better, cleaned, and easy to read. 
Good job.

ziemsky:
`6b4f3d9` rebases the branch onto latest master with no further changes.

PR #1473, opened 2021-05-21
ADZ-895 indicate level of nesting in schema with tree diagram

harryhmc:
Changes to implement rendering of a tree diagram alongside the schema, to further indicate nesting. Said tree diagram is made up of absolutely-positioned divs, referred to in code as "borders". The tree diagram ties in visually to expander/collapser buttons, which have been moved to the left of the name column. Each indented level has lighting and padding set as CSS vars.

image

The main bit of work is in introducing the BorderHelper. I have also updated the UUIDHelper to implement the Helper interface, and the IndentationHelper to let it return a weighted indentation value (useful for controlling the amount of padding between 'indentations').

### DIFF
--- a/cms/src/main/java/uk/nhs/digital/apispecs/handlebars/BorderHelper.java
+++ b/cms/src/main/java/uk/nhs/digital/apispecs/handlebars/BorderHelper.java
@@ -1,0 +1,239 @@
+package uk.nhs.digital.apispecs.handlebars;
+
+
+import com.github.jknack.handlebars.Handlebars.SafeString;
+import com.github.jknack.handlebars.Helper;
+import com.github.jknack.handlebars.Options;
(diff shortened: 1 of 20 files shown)


ziemsky (approved):
Excellent work :+1:

harryhmc:
I've since added some commits to deal with two bugs that were present in my original PR. These were:

1. Not accounting for certain nodes in the ContextModelStack being ArrayList<Schema>'s, rather than an extension of Schema
2. Not being able to distinguish between certain Schema nodes, due to property names not being set on Schema objects

The first was a relatively simple fix. The second has been fixed by adding preprocessing jobs to `ApiSpecificationStaticHtml2Codegen.java` for Request and Response schemas. These jobs add in the property name as a custom extension "x-property-name".

urchy (approved):
Well done! You smash it.

PR #1496, opened 2021-05-26
[ADZ-439] OAS path parameters propagation

ziemsky:
[ADZ-422] ensured that path parameters propagate
to operations where operations don't define
any of their own. It achieved that by overriding
a method of CodeGen's `DefaultGenerator`. The
latter is not designed for extensions, and
this required duplicating a lot of code in the
extending class to apply the fix.

It has now turned out, however, that the required
behaviour can be toggled through configuration,
and so this change applies the configuration
in place of the override (see the use of
`ParserConfiguration` in
`SwaggerCodeGenOpenApiSpecificationJsonToHtmlConverter`.

This change also:
* Upgrades Codegen components to their
  latest versions.
* Moves the change introduced by ADZ-992 from
  `CodegenDefaultGenerator` to
  `ApiSpecificationStaticHtml2Codegen`.
* Removes the now redundant `CodegenDefaultGenerator`
  class.
* Cleans up `SwaggerCodeGenOpenApiSpecificationJsonToHtmlConverter`
  making the flow more readable.

This is a re-introduction of `fb05e3b8` (reverting the
'revert' commit `fb05e3b8`).


urchy (approved):
Double check. Well done Rafal.

PR #1516, opened 2021-06-02
[ADZ-1178] First version working preview create and base changed

urchy:
Creating preview page with new templates - Header and Footer. 
Only accessible at api-catalogue/preview.


harryhmc (approved):
nice work m8

PR #1527, opened 2021-06-07
[DW-2410] Extra cookies removed from necessary only

mehranrashid:


repository-data/webfiles/src/main/resources/site/freemarker/common/global-footer.ftl
```diff
@@ -21,7 +21,7 @@
     <#if !hstRequest.requestContext.cmsRequest>
         <@ga.accountId/>
         <@hst.link var="googleAnalytics" path="/resources/google-analytics.js"/>
-        <script src="${googleAnalytics}"></script>
+        <script type="text/plain" data-cookieconsent="statistics" src="${googleAnalytics}"></script>
```
GeoffreyHayward, line 24: We learned it's not valid to have both a _type_ and _src_ on a _script_ element. When we looked into this in the passed we found that google-analytics.js does not set any cookies, until the second script tag it provides..

GeoffreyHayward: This does not seem to be true anymore, in terms of google not setting cookies at this point.

GeoffreyHayward (approved):

PR #1528, opened 2021-06-07
[ADZ-1180] Rebranding - document title banner

urchy:
Document title banner added to api-catalogue template.

### DIFF
--- a/repository-data/webfiles/src/main/resources/site/freemarker/common/api-catalogue-preview.ftl
+++ b/repository-data/webfiles/src/main/resources/site/freemarker/common/api-catalogue-preview.ftl
@@ -18,8 +18,11 @@
 <#-- Content Page Pixel -->
 <@contentPixel document.getCanonicalUUID() document.title></@contentPixel>
 
-<@headerBanner document />
+
 <article class="article article--filtered-list api-catalogue">
+
+    <@headerBanner document />
+
     <div class="grid-wrapper grid-wrapper--article">
         <div class="grid-row">
             <div class="column column--two-thirds column--reset">
--- a/repository-data/webfiles/src/main/resources/site/freemarker/common/api-catalogue-preview.ftl
+++ b/repository-data/webfiles/src/main/resources/site/freemarker/common/api-catalogue-preview.ftl
@@ -6,6 +6,7 @@
 <#include "macro/apiCatalogueEntries.ftl">
 <#include "macro/component/lastModified.ftl">
 <#include "macro/svgIcons.ftl">
+<#include "../nhsd-common/macros/header-banner.ftl">
 
 <#-- @ftlvariable name="document" type="uk.nhs.digital.website.beans.ComponentList" -->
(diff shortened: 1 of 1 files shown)


repository-data/webfiles/src/main/resources/site/freemarker/common/api-catalogue-preview.ftl
```diff
@@ -17,24 +18,16 @@
 <#-- Content Page Pixel -->
 <@contentPixel document.getCanonicalUUID() document.title></@contentPixel>
 
+<@headerBanner document />
```
harryhmc, line 21: comparing this to `general-article.ftl` suggests that the banner image should be _inside_ the article, just above the grid-wrapper div

repository-data/webfiles/src/main/resources/site/freemarker/common/api-catalogue-preview.ftl
```diff
@@ -17,24 +18,16 @@
...
-        </div>
-        APi Catalogue - Preview
         <div class="grid-row">
             <div class="column column--two-thirds column--reset">
                 <#-- [FTL-BEGIN] 'Summary and optional Body' sections -->
                 <div id="section-summary" class="article-section article-section--summary  no-border">
                     <div class="grid-row">
                         <div class="column column--reset">
                             <div class="rich-text-content">
-                                <div data-uipath="website.linkslist.summary"><@hst.html hippohtml=document.summary contentRewriter=gaContentRewriter/></div>
```
harryhmc, line 37: dont think i understand why this has been removed?

repository-data/webfiles/src/main/resources/site/freemarker/common/api-catalogue-preview.ftl
urchy: You right, I was taking a look at the list-article which has it outside of the article.
I have pushed another version with it inside the article.

repository-data/webfiles/src/main/resources/site/freemarker/common/api-catalogue-preview.ftl
urchy: This is now part of the headerBanner. its the same info

repository-data/webfiles/src/main/resources/site/freemarker/common/api-catalogue-preview.ftl
ziemsky: Good spot, Harry.

Having said that: I know I kept mentioning `general-article.ftl` as the main reference, but having looked more closely into this with Andre, it seems that `nhsd-list-article.ftl` is probably a better basis for API catalogue page.

In `nhsd-list-article.ftl` the `<@headerBanner document />` is not wrapped in any container element whilst in `general-article.ftl`. Based on this, if the page looks okay in testing with the header wrapped in another container element I think it ultimately doesn't matter much and we can go ahead with either variant.

ziemsky (approved):
:+1:

PR #1529, opened 2021-06-08
[ADZ-1219] OAS spec import and rerender fix

ziemsky:
Recent update of swagger-codegen-generators from
1.0.25 to 1.0.26 broke importing and re-rendering
of API specifications as the new version of the
library was compiled with newer Jackson version.

This commit fixes this by upgrading Jackson dependencies
to corresponding from version 2.11.2 to 2.12.2.

Also:
* In SchemaHelperTest restores a test case previously
  deleted by mistake.
* Makes SchemaHelperTest no longer create a temporary
  JSON datafile (feeding JSON to parser directly, instead).
* Expands test case
  rendersSpecification_whenCompleteOpenApiJsonAsHtml
  with data that triggered the (now fixed) bug.


urchy (approved):
Good job Rafal.

PR #1534, opened 2021-06-09
[DW-2450] External links to files use DownloadBlockAsset

thebib:


mehranrashid (approved):

thebib:
There's other uses of download block asset I need to check.

PR #1540, opened 2021-06-16
ADZ-1243 Updater Script to update nodes.

urchy:
We have implemented a solution to automate a manual process. 
The updater editor allows developers to run Groovy updater scripts against a running repository from within the CMS UI. This allows us to update the "frontend" code without restart the server. 

To test run it locally and go to: http://localhost:8080/cms/systemupdater 
Select rerenderScpecs from the list. Update two variables with your local paths and click execute.


ziemsky (approved):
:+1:

PR #1548, opened 2021-06-18
[ADZ-1183] ADZ-1183 final changes for MVP

harryhmc:
Changes for ADZ-1183 - rebranding the filter section
image

### DIFF
--- a/repository-data/webfiles/src/main/resources/site/freemarker/common/api-catalogue-preview.ftl
+++ b/repository-data/webfiles/src/main/resources/site/freemarker/common/api-catalogue-preview.ftl
@@ -21,7 +21,7 @@
 
 <@headerBanner document />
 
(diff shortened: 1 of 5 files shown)


ziemsky:
I'm approving in the interest of not holding back MVP go-live. We are aware that there are some elements in this change that are temporary and should be re-designed a bit (e.g. `_api-catalogue.scss` or  the couple of `*-newDesign.ftl` templates) but this is expected to happen very soon.

As part of this change the Filters section was styled using [tick-box based styling](https://design-system.digital.nhs.uk/storybook/latest/?path=/story/design-system-components-molecules-filter-menu-section--lab-component) but a new concept is being worked on and will be applied soon, so there is no point investing too much effort into polishing something up for it to shortly change anyway.

ziemsky (approved):

PR #1554, opened 2021-06-21
[DW-2150] News doctype rebranded

thebib:


thebib:
I have disabled the Blog.featue file, so the automation tests for this section will not run and there is a new ticket created to enable them and get them working again. This will be done in a separate PR.

mehranrashid (approved):

PR #1557, opened 2021-06-25
[ADZ-1185] API catalogue rebranded

ziemsky:
* Switches API catalogue to use the new
  re-branded styling, overwriting the
  old version with the 'preview' one
  where the rebranding work was done,
  and removes the temporary preview
  version.
* Groups API catalogue related
  templates together, to eliminate
  possibility of them being taken
  for being generic/reusable.
* Temporarily adds custom stylesheet
  until more 'proper' design of filters
  is defined and applied.
* Makes `az-nav.fts` accept list of
  links rather than the document object
  normally containing the links. This
  makes it possible to pre-filter the
  list of links passed to the template
  before it's being invoked, and makes
  the template suitable to be used
  in API catalogue where the list of
  links passed to A-Z template changes
  dynamically, depending on current
  user's choice.


repository-data/webfiles/src/main/resources/site/freemarker/common/apicatalogue/scrollable-filter-nav.ftl
```diff
@@ -0,0 +1,94 @@
...
+                    <input type="checkbox">${filter.displayName}
+                </a>
+                <span class="checkmark <#if filter.selected>selected</#if>"></span>
+            <#else>
+                <a href="" class="nhsd-a-checkbox__label nhsd-t-body-s nhsd-api-catalogue__filter-unavailable">
+                    <input type="checkbox" disabled>${filter.displayName}
+                </a>
+                <span class="checkmark <#if filter.selected>selected</#if>"></span>
+            </#if>
+                <p class="nhsd-a-checkbox__hint"></p>
```
LEJA3, line 84: Is this element required?

LEJA3 (commented):
Looks good. The only thing I noticed was a `<p>` tag with no content which you might want to remove if it's not needed.

repository-data/webfiles/src/main/resources/site/freemarker/common/apicatalogue/scrollable-filter-nav.ftl
ziemsky: Thanks, @LEJA3 !

It was placed here replicating example from the [Design System](https://design-system.digital.nhs.uk/storybook/latest/?path=/docs/design-system-components-molecules-filter-menu-section--lab-component).

I totally agree that it's superfluous in our case and I did try removing it before submitting this PR but it broke the layout (resulting in check-boxes and their labels being displayed in a single row), so I decided to leave it in.

LEJA3 (approved):

ziemsky:
`4716debd` is just a rebase onto latest `master`; no other changes compared to the reviewed `6b81522`.

PR #1603, opened 2021-07-14
[ADZ-1278] API catalogue redundant ids

ziemsky:


urchy (approved):
Good Job Rafal.
I like the way you crack the command to get those duplicate ids. Well done.

PR #1605, opened 2021-07-15
[ADZ-1192] API-Spec 'try this api' banner rebranding

urchy:
'Try this API' banner rebranding. 

Screenshot 2021-07-15 at 13 43 19

Update updater command `mkdir -p` to be compatible with both macOS and Linux


ziemsky (approved):
:+1:

PR #1621, opened 2021-07-27
[ADZ-1265] Deprecation of APIs - New Hide / Unhide feature

urchy:
The purpose of this implementation is to add a feature that hides any deprecated or retired APIs by default, but also gives the ability to unhide them again.
To manage this feature I have created a new HTTP request parameter called "showAll" to manage the toggle that allows the user to hide/unhide deprecated and retired specs.
This showAll flag manages by the user and is false by default has the responsibility to filter the entire list of specs the user has on the screen.

[ADZ-1265](https://nhsd-jira.digital.nhs.uk/browse/ADZ-1265)


repository-data/webfiles/src/main/resources/site/freemarker/common/apicatalogue/api-catalogue.ftl
```diff
@@ -20,10 +20,21 @@
...
+                <#if showAll??>
+                    <#if showAll == true>
+                        <#assign toShowAll="false">
+                    <#else>
+                        <#assign toShowAll="true">
+                    </#if>
+                </#if>
+
+                <@hst.renderURL var="showAlllink">
+                    <@hst.param name="showAll" value="${toShowAll}" />
```
ziemsky, line 35: If you replace:
```
${toShawAll}
```
with:
```
${showAll?string('false', 'true')}
```
you'll be able to get rid of the whole `<# if showAll??> ... </#if>` construct and the `toShowAll` variable.

repository-data/webfiles/src/main/resources/site/freemarker/common/apicatalogue/api-catalogue.ftl
```diff
@@ -20,10 +20,21 @@
...
             <#if document.body?has_content??>
+                <#if showAll??>
+                    <#if showAll == true>
+                        <#assign toShowAll="false">
+                    <#else>
+                        <#assign toShowAll="true">
+                    </#if>
+                </#if>
+
+                <@hst.renderURL var="showAlllink">
```
ziemsky, line 34: Would be good to move the link generation closer to (right before) the place where it's being used.

ziemsky: The `showAlllink` wants to be capitalised as `showAllLink` I think.

site/components/src/main/java/uk/nhs/digital/common/components/apicatalogue/ApiCatalogueComponent.java
```diff
@@ -56,11 +62,33 @@ private Session sessionFrom(final HstRequest request) {
...
+
+    private List<Internallink> removeDeprecatedAndRetiredLinks(HstRequest request) {
+
+        List<?> blocks = ((ComponentList) request.getRequestContext().getContentBean()).getBlocks();
+
+        return blocks.stream()
+            .filter(Internallink.class::isInstance)
+            .map(Internallink.class::cast)
+            .filter(link -> !linkHasSelectedTaxonomyKeys(link, ImmutableSet.of("deprecated")))
+            .filter(link -> !linkHasSelectedTaxonomyKeys(link, ImmutableSet.of("retired")))
```
ziemsky, line 86: Given that we have two keys to be excluded, how about passing them as a set and invoke the method just once? After all, it already supports a collection.

site/components/src/main/java/uk/nhs/digital/common/components/apicatalogue/ApiCatalogueComponent.java
```diff
@@ -56,11 +62,33 @@ private Session sessionFrom(final HstRequest request) {
             .orElse(Collections.emptySet());
     }
 
-    private List<?> apiCatalogueLinksFrom(final HstRequest request) {
-        return ((ComponentList) request.getRequestContext().getContentBean()).getBlocks();
+    private List<Internallink> apiCatalogueLinksFrom(final HstRequest request, final Boolean isToShowAll) {
+
+        return isToShowAll ? retrieveAllLinks(request) : removeDeprecatedAndRetiredLinks(request);
```
ziemsky, line 67: It feels that it would read easier if we could avoid this conditional and avoid passing the `request` object around, instead extracting the collection of links once and filter it in stages.

How about moving this logic to `apiCatalogueLinksFilteredBySelectedTags`?

The `apiCatalogueLinksFilteredBySelectedTags` would then look something like this:
```
    private List<Internallink> apiCatalogueLinksFilteredBySelectedTags(final List<Internallink> links, final Set<String> selectedTags, final boolean showAll) {

        final List<Internallink> linksWithoutRetiredOrDeprecated = links.stream()
            .filter(link -> showAll || !linkHasSelectedTaxonomyKeys(link, ImmutableSet.of("deprecated-api","retired-api")));
        
        if (selectedTags.isEmpty()) {
            return linksWithoutRetiredOrDeprecated.collect(toList());
        }

        return internalLinksWithSelectedTaxonomyKeys(linksWithoutRetiredOrDeprecated, selectedTags).collect(toList());
    }
```

The name of the method should then probably change to simply say 'filtered' without qualifying 'by what', e.g. by dropping the `BySelectedTags` bit. Naturally, this needs testing to see if the change does not mess up the logic.

Notice that I also changed the key names assuming that my proposal communicated on Slack will be accepted.

ziemsky (approved):
Nice work!

acceptance-tests/src/test/resources/features/site/apiCatalogue.feature
```diff
@@ -1,3 +1,4 @@
+@WIP
```
ziemsky, line 1: A minor leftover :grin:

PR #1622, opened 2021-07-27
[ADZ-769] ADZ-769 implementing mechanism for ordering endpoint headings

harryhmc:
Re-introduced CodegenDefaultGenerator.java to allow for custom ordering of CodegenOperations. annoyingly, processOperation is a private method so the generator can't just call the super method

### DIFF
--- a/cms/src/main/java/uk/nhs/digital/apispecs/swagger/CodegenDefaultGenerator.java
+++ b/cms/src/main/java/uk/nhs/digital/apispecs/swagger/CodegenDefaultGenerator.java
@@ -1,0 +1,51 @@
+package uk.nhs.digital.apispecs.swagger;
+
+import static java.util.Collections.singletonList;
+import static java.util.stream.Collectors.toList;
+
+import io.swagger.codegen.v3.CodegenOperation;
+import io.swagger.codegen.v3.DefaultGenerator;
+import io.swagger.v3.oas.models.Paths;
+import io.swagger.v3.oas.models.tags.Tag;
+
+import java.util.*;
+
+/**
+ * Some producer teams require ordering their endpoints in a non-lexographical order.
+ * This requires a custom extension of the DefaultGenerator provided by CodeGen.
+ *
+ * This class CodegenDefaultGenerator extends DefaultGenerator class and sorts CodegenOperations before returning them.
+ * See the method orderedByDeclaredTags.
+ */
+public class CodegenDefaultGenerator extends DefaultGenerator {
+
+    @Override
+    public Map<String, List<CodegenOperation>> processPaths(Paths paths) {
+        Map<String, List<CodegenOperation>> ops = super.processPaths(paths);
+        return orderedByDeclaredTags(ops);
+    }
+
+    private Map<String, List<CodegenOperation>> orderedByDeclaredTags(final Map<String, List<CodegenOperation>> operationsByTags) {
+        final List<String> declaredTagNames = operationsByTags.containsKey("Default")
+            ? new ArrayList<>(singletonList("Default"))
+            : new ArrayList<>(operationsByTags.size());
+
+        declaredTagNames.addAll(
+            Optional.ofNullable(openAPI.getTags())
+                .orElse(Collections.emptyList())
+                .stream()
+                .map(Tag::getName)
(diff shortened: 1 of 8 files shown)


cms/src/main/java/uk/nhs/digital/apispecs/swagger/CodegenDefaultGenerator.java
```diff
@@ -0,0 +1,194 @@
...
+
+        declaredTagNames.stream()
+            .filter(operationsByTags::containsKey)
+            .forEachOrdered(tagName -> orderedOperationsByTags.put(tagName, operationsByTags.get(tagName)));
+
+        return orderedOperationsByTags;
+    }
+
+    @Override
+    public List<File> generate() {
```
ziemsky, line 189: I know this override was included in my example patch but I believe it can actually go.

The patch was generated before the fix implemented by this override was replaced by `ApiSpecificationStaticHtml2Codegen#populateComponentsFieldWithEmptyObjectWhenNull`.

cms/src/main/java/uk/nhs/digital/apispecs/swagger/CodegenDefaultGenerator.java
```diff
@@ -0,0 +1,194 @@
...
+ * Some producer teams require ordering their endpoints in a non-lexographical order.
+ * This requires a custom extension of the DefaultGenerator provided by CodeGen.
+ *
+ * This class CodegenDefaultGenerator extends DefaultGenerator class and sorts CodegenOperations before returning them.
+ * See the method orderedByDeclaredTags.
+ */
+public class CodegenDefaultGenerator extends DefaultGenerator {
+
+    @Override
+    public Map<String, List<CodegenOperation>> processPaths(Paths paths) {
```
ziemsky, line 32: How about replacing this method (and the chain of duplicated private ones with something like this?:
```
    @Override
    public Map<String, List<CodegenOperation>> processPaths(Paths paths) {
        return orderedByDeclaredTags(super.processPaths(paths));
    }
```

harryhmc: oh that's much better! i've slightly rewritten that just to avoid a bunch of nested method calls
```
@Override
public Map<String, List<CodegenOperation>> processPaths(Paths paths) {
    Map<String, List<CodegenOperation>> ops = super.processPaths(paths);
    return orderedByDeclaredTags(ops);
}
```
i say "a bunch" - i realise this only avoids one whole level of nesting :sweat_smile:

ziemsky (approved):
:+1:

cms/src/main/java/uk/nhs/digital/apispecs/swagger/CodegenDefaultGenerator.java
```diff
@@ -0,0 +1,51 @@
+package uk.nhs.digital.apispecs.swagger;
+
+import static freemarker.template.utility.Collections12.singletonList;
```
ziemsky, line 3: I've only just noticed: this wants to be `import static java.util.Collections.singletonList;`

cms/src/main/java/uk/nhs/digital/apispecs/swagger/CodegenDefaultGenerator.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.*;
+
+/**
+ * Some producer teams require ordering their endpoints in a non-lexographical order.
+ * This requires a custom extension of the DefaultGenerator provided by CodeGen.
+ *
+ * This class CodegenDefaultGenerator extends DefaultGenerator class and sorts CodegenOperations before returning them.
+ * See the method orderedByDeclaredTags.
+ */
+public class CodegenDefaultGenerator extends DefaultGenerator {
```
ziemsky, line 20: Would you mind adding this class to the compile list in `ApiSpecificationRenderer.groovy` and make sure the updater script works, Harry?

harryhmc: yep, compiles without issues

PR #1651, opened 2021-08-09
[ADZ-1170] Try this API: request body default example fix

ziemsky:
Fixes the issue by upgrading Swagger UI to version 3.51.2.

Also:
* Enables scroll bar in the curl command box.
* Makes `tryThisApiOauth2RedirectUrl` being generated
  on the server rather than through JavaScript.
* Eliminates warning *Helper 'variable' has been replaced...*
  from the logs, but removing redundant registration of
  one of the Handlebars helpers.


urchy (approved):
Good job. Simple and clean.

ziemsky:
`8fc49c1` is just a rebase onto the latest `master` with no functional changes compared to the reviewed `61a2dcb`.

PR #1673, opened 2021-08-16
[ADZ-898] API spec: Example vs Allowed Values

ziemsky:


harryhmc (commented):
LGTM, great work around improving the test suite :+1:

harryhmc (approved):

PR #1675, opened 2021-08-16
[ADZ-1312] API-Description - Add horizontal line above <h2> headers

urchy:
This user story has changed the way we parse the description text of specifications. 

Until now we just manage headers size from markdown. Based on the number of hashes(#) we set a number of attributes implementing `AttributeProvider` interface. However, it has some limitations that don't allow us to have proper control to add new HTML elements like `<hr>`

In order to have full control, I have implemented `NodeRenderer` interface, another interface that allows full control of nodes.

Within these changes I have deleted `HeadingAttributeProvider` class that was an `AttributeProvider` implementation and use the same logic in a new class `HeadingsRenderer` that implements a `NodeRenderer`. Now we have the same logic as before plus the possibility to add new HTML elements.

### DIFF
--- a/cms/src/main/java/uk/nhs/digital/apispecs/commonmark/CommonmarkMarkdownConverter.java
+++ b/cms/src/main/java/uk/nhs/digital/apispecs/commonmark/CommonmarkMarkdownConverter.java
@@ -9,11 +9,13 @@ import org.commonmark.ext.gfm.tables.TablesExtension;
 import org.commonmark.node.AbstractVisitor;
 import org.commonmark.node.Heading;
 import org.commonmark.node.Node;
+import org.commonmark.node.ThematicBreak;
 import org.commonmark.parser.Parser;
 import org.commonmark.renderer.html.HtmlRenderer;
 
 import java.util.List;
 import java.util.Optional;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 
 public class CommonmarkMarkdownConverter {
@@ -131,17 +133,20 @@ public class CommonmarkMarkdownConverter {
 
         normalizeHeadingsLevels(document, topHeadingLevel);
 
+        addBreaksBeforeH2Headings(document);
+
         return renderMarkdownModel(document, extensions, headingIdPrefix);
     }
 
     private String renderMarkdownModel(final Node document, final List<Extension> extensions, final String headingIdPrefix) {
 
         final HtmlRenderer renderer = HtmlRenderer.builder()
-            .nodeRendererFactory(context -> new FencedCodeBlockNodeRenderer(context))
-            .nodeRendererFactory(context -> new TableBlockNodeRenderer(context))
+            .nodeRendererFactory(FencedCodeBlockNodeRenderer::new)
+            .nodeRendererFactory(TableBlockNodeRenderer::new)
             .attributeProviderFactory(context -> new CodeAttributeProvider())
-            .attributeProviderFactory(context -> new HeadingAttributeProvider(headingIdPrefix))
             .attributeProviderFactory(context -> new ListAttributeProvider())
+            .attributeProviderFactory(context -> new ThematicBreakAttributeProvider())
+            .attributeProviderFactory(context -> new HeadingAttributeProvider(headingIdPrefix))
             .attributeProviderFactory(context -> new HyperlinkAttributeProvider())
             .attributeProviderFactory(context -> new ParagraphAttributeProvider())
             .attributeProviderFactory(context -> new StrongEmphasisAttributeProvider())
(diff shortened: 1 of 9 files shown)


cms/src/main/java/uk/nhs/digital/apispecs/commonmark/CommonmarkMarkdownConverter.java
```diff
@@ -137,10 +137,10 @@ private String renderMarkdownText(final String markdown, final String headingIdP
     private String renderMarkdownModel(final Node document, final List<Extension> extensions, final String headingIdPrefix) {
 
         final HtmlRenderer renderer = HtmlRenderer.builder()
-            .nodeRendererFactory(context -> new FencedCodeBlockNodeRenderer(context))
-            .nodeRendererFactory(context -> new TableBlockNodeRenderer(context))
+            .nodeRendererFactory(FencedCodeBlockNodeRenderer::new)
+            .nodeRendererFactory(TableBlockNodeRenderer::new)
+            .nodeRendererFactory(context -> new HeadingsRenderer(context, headingIdPrefix))
```
harryhmc, line 142: super pedantic, but could this class be renamed `HeadingNodeRenderer` to match it's siblings?

cms/src/test/java/uk/nhs/digital/apispecs/CommonmarkMarkdownConverterTest.java
```diff
@@ -141,66 +121,25 @@ public void rendersHeadingIds_addingSuffixes_whenDuplicateHeadingValues() {
...
-        final Levels initialHeadingLevels,
-        final Levels expectedHeadingLevels
-    ) {
+    public void rendersHeadings_withHeadingsHierarchyLevelsAdjustedViaParameter() {
         // given
-        final String markdown = markdownWithHeadingsAt(initialHeadingLevels);
+        final String markdown = from("headingsWithHierarchyLevels.md");
 
-        final String expectedHtml = htmlWithHeadingsAt(expectedHeadingLevels);
+        final String expectedHtml = from("headingsWithHierarchyLevels.html");
```
harryhmc, line 128: i'm guessing you've made the move to a static test file to get around issues with \<hr\> elements throwing off your assertions - i think this makes the test weaker since it now covers much less cases

you could always strip the hr elements from the text before comparing them, similarly to how we strip whitespace in other methods? and write a test method specifically for checking that hr's are inserted where we expect them

cms/src/main/java/uk/nhs/digital/apispecs/commonmark/CommonmarkMarkdownConverter.java
harryhmc: also i like the use of the method reference operator for `FencedCodeBlockNodeRenderer` and `TableBlockNodeRenderer` :grin:

cms/src/test/java/uk/nhs/digital/apispecs/CommonmarkMarkdownConverterTest.java
```diff
@@ -241,8 +234,11 @@ private String htmlWithHeadingsAt(final Levels headingsLevels) {
         // <h3 id="heading">Heading</h3>
 
         return headingsLevels.stream()
-            .map(level -> format("<h{0} class=\"{1}\">Heading</h{0}>", level, cssClassForHeadingLevel(level)))
+            .map(level -> level == 2 ? format("<hr class=\"nhsd-a-horizontal-rule\">\n")
```
ziemsky, line 237: :+1:

cms/src/test/java/uk/nhs/digital/apispecs/commonmark/ThematicBreakAttributeProviderTest.java
```diff
@@ -0,0 +1,54 @@
...
+
+    private ThematicBreakAttributeProvider thematicBreakAttributeProvider = new ThematicBreakAttributeProvider();
+
+    @Before
+    public void setUp() {
+        initMocks(this);
+    }
+
+    @Test
+    public void setsCodeInlineCssClassOnCodeElement() {
```
ziemsky, line 29: The method's name was probably meant to reflect the fact that we're dealing with the 'thematic break' line, here :wink:

cms/src/test/java/uk/nhs/digital/apispecs/commonmark/ThematicBreakAttributeProviderTest.java
```diff
@@ -0,0 +1,54 @@
...
+        // when
+        thematicBreakAttributeProvider.setAttributes(nodeCode, "tagName is ignored", attributes);
+
+        // then
+        then(attributes).should().put("class", "nhsd-a-horizontal-rule");
+        then(attributes).shouldHaveNoMoreInteractions();
+    }
+
+    @Test
+    public void ignoresElementOtherThanCode() {
```
ziemsky, line 43: Similar comment regarding naming to the one above.

cms/src/test/java/uk/nhs/digital/apispecs/commonmark/ThematicBreakAttributeProviderTest.java
```diff
@@ -0,0 +1,54 @@
...
+
+        // given
+        final Node nodeOtherThanCode = new TableBlock();
```
ziemsky, line 46: Similar comment regarding naming to the one above.

ziemsky (approved):
:+1: 

Nice work, Andre. Couple of cosmetic comments regarding some names in one of the tests, presumably carried over from an example test.

cms/src/test/java/uk/nhs/digital/apispecs/CommonmarkMarkdownConverterTest.java
harryhmc: can we remove the `format` call from this line as it isn't doing anything?

ziemsky: Good spot, Harry. Sorted.

PR #1687, opened 2021-08-31
[ADZ-1353] ADZ-1353 split sidenav into two boxes

harryhmc:
Freemarker template changes to split sidenav into AZ block and filters block


ziemsky (approved):
Nice one :+1:

PR #1705, opened 2021-09-07
Adz 1161 directory maven plugin fix

ziemsky:


ziemsky:
Implemented as #1706.

PR #1706, opened 2021-09-07
[DW-0000] Directory Maven plugin workaround

ziemsky:
Latest Maven version breaks directory-maven-plugin.

The plugin is used to identify filesystem path of the
root module of the Maven project, which is then exposed
to the other modules via `website.basedir` property
(note that `project.basedir` property changes between
modules and so is not suitable).

The workaround is to apply property
`maven.multiModuleProjectDirectory` instead of the
plugin and instead of `website.basedir`.

This is only a temporary workaround, because
`maven.multiModuleProjectDirectory` is Maven's internal,
undocumented property which may change without warning
in Maven's future versions and so its use is discouraged.

A proper fix to the issue is in progress of being developed,
and once
https://github.com/jdcasey/directory-maven-plugin/issues/16
issue is addressed, the plugin should be updated and the current
workaround reversed.

### DIFF
--- a/acceptance-tests/pom.xml
+++ b/acceptance-tests/pom.xml
@@ -357,36 +357,36 @@
                             <configuration>
                                 <configfiles>
                                     <configfile>
-                                        <file>${website.basedir}/conf/context.xml</file>
+                                        <file>${maven.multiModuleProjectDirectory}/conf/context.xml</file>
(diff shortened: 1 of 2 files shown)


GeoffreyHayward (approved):
Thank you.

PR #1715, opened 2021-09-09
[DW-1823] Update repo to reflect maven issue

lekeDevOps:
update to reflect maven workaround.


ziemsky (approved):

ziemsky:
The only change in this PR is a removal of a single, empty line from log config file.

I approved on the grounds that there can be no possible harm from this change, and under assumption that this is just to trigger some event in Splunk.

PR #1724, opened 2021-09-14
[ADZ-1358] filter refinement

harryhmc:
Updating the filters in the sidenav

- Increased vertical spacing between filters
- Re-implemented checkbox border on hover, reduced border size
- Changed filter text colour

image

### DIFF
--- a/repository-data/webfiles/src/main/resources/site/freemarker/common/apicatalogue/scrollable-filter-nav.ftl
+++ b/repository-data/webfiles/src/main/resources/site/freemarker/common/apicatalogue/scrollable-filter-nav.ftl
@@ -10,7 +10,7 @@
         <h2 id="side-az-nav-heading-responsive" class="nhsd-t-heading-xs">Search A-Z</h2>
         <@azList blockGroups "side-az-nav-heading-responsive"/>
     <#else>
-        <h2 id="side-az-nav-heading" class="nhsd-t-heading-xs">Refine results</h2>
+        <h2 id="side-az-nav-heading" class="nhsd-t-heading-xs">Search A-Z</h2>
         <@azList blockGroups "side-az-nav-heading"/>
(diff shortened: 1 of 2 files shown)


ziemsky (approved):
:+1:

repository-data/webfiles/src/main/resources/site/freemarker/common/apicatalogue/scrollable-filter-nav.ftl
```diff
@@ -80,6 +80,7 @@
         <#else>
             <#assign filtersParam = filtersModel.selectedFiltersKeysPlus(filter.key) />
         </#if>
+        <div>
```
ziemsky, line 83: It would be good to add an explaining comment if this div is needed, otherwise it may be removed by accident (which, I'm guessing, is what removed the tick box highlights in the past?).

Using Freemarker comment syntax, so that it doesn't get rendered in the HTML, this would look something like:

`<div> <#-- The reason for the div here -->`

I'm guessing this div is needed to enable the highlights around tick boxes?

ziemsky: Also: not related to your change, but I've just spotted it - A-Z title in `scrollable-filter-nav.ftl` is 'Refine filters' for desktop view and 'Search A-Z' for respoinsive; would be good to align. I think the 'Search A-Z' is later and so it should win.

harryhmc: Nope, the div is actually for adding vertical spacing between filters. It doesn't work if you just add spacing properties to the span, i think because it's considered an inline HTML element

Added a comment and updated the responsive A-Z title!

PR #1735, opened 2021-09-22
[ADZ-1365] Slimmer Try this API banner

SatnamSinghUK:
Also simplifies a couple of tests,
removing redundant test data files
in the process.


ziemsky (approved):
:+1:

PR #1761, opened 2021-10-04
[ADZ-1387] ADZ-1387 fixing cursor hover

harryhmc:
* also fixes issue with filter checkboxes in chrome
* also fixes issues with disabled filters being interactive


ziemsky (approved):
Nice work, Harry :+1:

PR #1776, opened 2021-10-14
Adz 1103 api catalogue filters doc links

SatnamSinghUK:


repository-data/webfiles/src/main/resources/site/freemarker/common/apicatalogue/scrollable-filter-nav.ftl
```diff
@@ -82,32 +84,40 @@
...
                 <a class="nhsd-a-checkbox__label nhsd-t-body-s nhsd-api-catalogue__filter-unavailable <#if filter.entries?has_content>nhsd-!t-font-weight-bold</#if>">
-                    <input type="checkbox" disabled>&nbsp;${filter.displayName}
+                    <input type="checkbox" disabled>
+                    <span class="<#if !responsive>filter-label__text</#if>">${filter.displayName}</span>
                 </a>
                 <span class="checkmark <#if filter.selected>selected</#if>"></span>
             </#if>
-                <p class="nhsd-a-checkbox__hint"></p>
             </label>
+            <@filterDescription responsive filter.description()/>
```
harryhmc, line 104: since the filterDescription is rendered _outside_ of the \<a\> element, am i right thinking it isn't actually "clickable"? in this case it might be better for the cursor to not be a pointer when hovering over it

repository-data/webfiles/src/main/resources/site/freemarker/common/apicatalogue/scrollable-filter-nav.ftl
```diff
@@ -5,6 +5,7 @@
 <#assign lettersOfTheAlphabet = ["A", "B", "C", "D", "E", "F", "G", "H", "I", "J", "K", "L", "M", "N", "O", "P", "Q", "R", "S", "T", "U", "V", "W", "X", "Y", "Z"]/>
 
 <#macro scrollableFilterNav blockGroups filtersModel responsive>
+    <#-- @ftlvariable name="filtersModel" type="uk.nhs.digital.common.components.apicatalogue.filters.Filters" -->
```
harryhmc, line 8: could you remind me why we're redeclaring the variables here now?

cms/src/main/java/uk/nhs/digital/common/validators/NegativeRegExpValidator.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    @Override
+    public Optional<Violation> validate(final ValidationContext context, final String text) {
+        return violationIfAnyForbiddenPatternsPresentInText(context, text);
+    }
+
+    private Optional<Violation> violationIfAnyForbiddenPatternsPresentInText(final ValidationContext context, final String text) {
+        return forbiddenPatterns.stream()
+            .filter(forbiddenWord -> containedIn(forbiddenWord, text))
```
harryhmc, line 55: personally i'd have used a `Predicate` like
```
Predicate<Pattern> isContainedInText = forbiddenWord -> forbiddenWord.match(text).find();

return forbiddenPatterns.stream()
    .filter(isContainedInText)
```
but that's just a difference in style really. your version is probably more performant since it doesn't rely on a new Predicate object per validation attempt

repository-data/application/src/main/resources/hcm-config/configuration/modules/validation.yaml
```diff
@@ -9,6 +9,14 @@ definitions:
       hipposys:className: org.onehippo.cms.services.validation.validator.RegExpValidator
       jcr:primaryType: hipposys:moduleconfig
       regexp.pattern: \d{4}-\d{4}
+    /hippo:configuration/hippo:modules/validation/hippo:moduleconfig/no-script:
+      forbiddenPatterns:
+      - \<\s*\bscript\b
```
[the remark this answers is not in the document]
ziemsky: Excellent find and observations! Thanks for that.

I think I'll drop support for HTML and instead implement just a support for hyperlink and paragraph in Markdown style. Feels safer than the current whitelisting approach.

PR #1777, opened 2021-10-15
[ADZ-1386] ADZ-1386 side-nav refinement

harryhmc:
* reduced filter menu max-height
* added margin-bottom to api catalogue content


ziemsky (approved):
You've made it a quick work, Harry :+1:

PR #1804, opened 2021-11-03
[ADZ-1495] Exclude API spec HTML from Lucene

ziemsky:
All properties of documents of type `website:apispecification` used to be indexed by default.

This change limits indexing on local machines only to relevant ones (actual goal: exclude `website:html` one from indexing).


Vlets (commented):
Changes look good, nothing to comment, needs to be tested on an environment.

Vlets (approved):

ziemsky:
`13a9024` is just a rebase onto latest `master`.

PR #1841, opened 2021-11-25
Dw 2335 email monitoring

lekeDevOps:
Hi both, would either be kind enough to review this? KR


cms/src/main/java/uk/nhs/digital/EmailMonitoringComponent.java
```diff
@@ -0,0 +1,47 @@
...
+                to[0] = new InternetAddress("emailmonitoring@nhs.net");
+                message.setRecipients(Message.RecipientType.TO, to);
+                message.setSubject("EMAIL MONITOR ONLY");
+                message.setContent("AUTOMATED", "text/plain");
+                Transport.send(message);
+                log.info("EMAIL MONITOR: SUCCESS");
+            } catch (MessagingException mex) {
+                mex.printStackTrace();
+                log.warn("EMAIL MONITOR: FAILED");
+                log.warn(mex.getMessage());
```
Deven-Rawat, line 41: Cloud you please change this to :  
` log.error("EMAIL MONITOR: FAILED ", MEX);`
Or you want to ignore this with warning?

cms/src/main/java/uk/nhs/digital/EmailMonitoringComponent.java
```diff
@@ -0,0 +1,47 @@
...
+            }
+        } catch (NamingException e) {
+            log.warn(e.getMessage());
```
Deven-Rawat, line 44: Do you want to ignore the exception with warning?

cms/src/main/java/uk/nhs/digital/EmailMonitoringComponent.java
lekeDevOps: Good point, this would be excellent for further debugging. It is only a few lines of code, that is quite simple. The main purpose is we get the trimmed exception message which should give us some indication without putting out too much information to the log. This is why the getMessage() was selected purely for simplicity.

cms/src/main/java/uk/nhs/digital/EmailMonitoringComponent.java
lekeDevOps: Good point, this would be excellent for further debugging. It is only a few lines of code, that is quite simple. The main purpose is we get the trimmed exception message which should give us some indication without putting out too much information to the log. This is why the getMessage was selected purely for simplicity.

Deven-Rawat (approved):

PR #1908, opened 2022-01-09
[ADZ-1433] API specs rendered on request

ziemsky:
Logic that renders API Specifications' HTML
from JSON, has been moved to the `site` module.

It is now executed when page is requested,
rather than during import of the specification's
JSON from Apigee.

Rendering is triggered by
`ApiSpecificationRendererDirective` invoked
from `apispecification.ftl` template.

SwaggerCodeGenOpenApiSpecificationJsonToHtmlConverter
is now defined as Spring bean.

ApiSpecRerenderJob and associated re-rendering
logic has been removed.


site/components/src/main/java/uk/nhs/digital/common/components/apispecification/ApiSpecificationRendererDirective.java
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+        try {
+            log.debug("Rendering API specification {}: start", path);
+            final String specificationHtml = converter.htmlFrom(specificationJson);
+            log.debug("Rendering API specification {}: done", path);
+
+            environment.getOut().append(specificationHtml);
+
+        } catch (final RuntimeException | IOException e) {
```
harryhmc, line 43: i've never seen the bitwise OR operator used in this way, could you explain how it works? moreso i can use it in the future, rather than understand it here

harryhmc (approved):
Great work Rafal - very grateful I had a walkthrough of this ahead of time!

site/components/src/main/java/uk/nhs/digital/common/components/apispecification/ApiSpecificationRendererDirective.java
ziemsky: Keen eye!

It's a shorthand for [handling exceptions of various types in a single `catch` clause](https://docs.oracle.com/javase/7/docs/technotes/guides/language/catch-multiple.html), rather than having one catch clause per exception type.

harryhmc: now here i was hoping i'd somehow missed [union types](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#union-types) being introduced to java :joy:

ziemsky:
Closing without merging, as this change depends on page caching being available, and will be released together with ADZ-1547 which provides it.

It has already been incorporated into the `ADZ-1547_custom_api_spec_page_caching` branch; pull request #1936.

PR #1936, opened 2022-01-19
ADZ-1547 Custom api spec page caching

ziemsky:
Only the ADZ-1547 commit needs reviewing.

ADZ-1433 has been reviewed as part of PR #1908.


repository-data/application/src/main/resources/hcm-content/content/documents/administration/admin-command-interface.yaml
```diff
@@ -0,0 +1,43 @@
...
+    hippostdpubwf:createdBy: admin
+    hippostdpubwf:creationDate: 2022-01-18T20:13:48.231Z
+    hippostdpubwf:lastModificationDate: 2022-01-18T20:13:48.231Z
+    hippostdpubwf:lastModifiedBy: admin
+    hippotranslation:id: cc919a10-e73d-4370-a8ec-88c6b2d40ff9
+    hippotranslation:locale: document-type-locale
+    jcr:mixinTypes:
+    - mix:referenceable
+    jcr:primaryType: common:stringlist
+  /admin-command-interface[2]:
```
Vlets, line 20: @ziemsky I would keep this one in mind in case something goes wrong. I'm unsure if get(0) will always be safe here.

site/components/src/main/java/uk/nhs/digital/admin/AdminCommand.java
```diff
@@ -0,0 +1,69 @@
...
+
+        return new EqualsBuilder().append(keyword, that.keyword).append(arguments, that.arguments).isEquals();
+    }
+
+    @Override public int hashCode() {
+        return new HashCodeBuilder(17, 37).append(keyword).append(arguments).toHashCode();
+    }
+
+    @Override public String toString() {
+        return new ToStringBuilder(this, SHORT_PREFIX_STYLE)
```
Vlets, line 64: @ziemsky I might be wrong as my understanding of the innards of a string builder is a bit hazy, but I'm unsure of thread safety here. 
What would happen if multithreading were to go wrong and send a duplicate command or a string with repeated keyword/arguments? I believe you mentioned there is a check for invalid arguments, however it could be a potential vulnerability for a diligent attacker.

repository-data/application/src/main/resources/hcm-content/content/documents/administration/admin-command-interface.yaml
ziemsky: Will do. Thanks for the warning, @Vlets :+1: 

So far, in my tests, there was always exactly one sub-node returned as long as the document has a published version. If the document is not published, the '`get(0)` invoking method will not even be called. Should the invocation ever be done in a different context, an error will be logged with a full stack trace which, hopefully, should be enough for a developer to understand what went wrong on that rare occasion when this feature is at all used.

Having said, I'll have a think and will consider implementing some cleaner way of dealing with this potential situation.

site/components/src/main/java/uk/nhs/digital/admin/AdminCommand.java
ziemsky: Conversation taken offline.

repository-data/application/src/main/resources/hcm-content/content/documents/administration/admin-command-interface.yaml
ziemsky: Fixed: `AdminCommandProvider` is now more resilient to situations where the structure of the Admin Command Interface document is not _exactly_ as it requires.

ziemsky:
`ff55536` applies changes recommended in code review (see comment about `AdminCommandProvider` above) and rebases the branch onto latest `master`.

This is the only commit to review; `fe39d6d` has been reviewed in a separate code review, on a separate branch.

GeoffreyHayward (approved):
Super work.

ziemsky:
The force-pushed `8f492f2` is just a rebase onto latest `master`, and includes no functional changes to the already reviewed code.

PR #1958, opened 2022-01-27
[DW-2740] Changed code for accessibility

Deven-Rawat:


LEJA3:
Closing this PR as the chart is accessible, but some charts contain invalid HTML which will be resolved in, DW-2885

PR #1963, opened 2022-01-31
[DW-000] Fixed comment style

LEJA3:


John-Milne (approved):
The correction is okay

John-Milne (approved):
Change allows for older documents where headingLevel does not exist

PR #2080, opened 2022-03-16
[DW-2882] Change to media thumbnail from content

lekeDevOps:
hi @Deven-Rawat would you do the honours?

KR
Lesley


site/components/src/main/java/uk/nhs/digital/common/feed/RssModifier.java
```diff
@@ -15,7 +15,7 @@
 import org.hippoecm.hst.core.request.HstRequestContext;
 import org.hippoecm.repository.util.DateTools;
 import org.jdom2.Element;
-import org.jdom2.Namespace;
+//import org.jdom2.Namespace;
```
Deven-Rawat, line 18: Please remove the commented import statement

lekeDevOps: Of course, not a problem. I was not too sure who commented it out and that maybe there was a reason for it, so did not want to remove it.

Deven-Rawat (approved):

PR #2083, opened 2022-03-18
[DW-2934] Script added to remove config files

Deven-Rawat:


.github/workflows/remove-old-configuration-files.yml
```diff
@@ -0,0 +1,110 @@
+# This is a basic workflow to help you get started with Actions
+
```
GeoffreyHayward, line 2: Please remove this comment and the extra line.

.github/workflows/remove-old-configuration-files.yml
```diff
@@ -0,0 +1,110 @@
...
+          CODE: ${{ steps.jwt_token_verify.outputs.code }}
+
+      - name: Get All Configuration Files
+        id: all-config-file-ids
+        run: |
+          response=$(curl \
+          -H "Authorization: Bearer ${{ steps.jwt_token.outputs.token1 }}" \
+          -X GET https://api.${{ secrets.HOST }}/v3/appconfigfiles)
+          echo "::set-output name=allConfigFiles::${response}"
+          echo "Files are " ${response}
```
GeoffreyHayward, line 60: This line can probably go. If anyone needs to debug in the future, it will be printed using the statement on lines 77 and 78 in the GitHub Actions log.

GeoffreyHayward (changes requested):
Looks good. Just 2 changes. Also what line(s) does it have the condition that is all but the latest last 3?

.github/workflows/remove-old-configuration-files.yml
Deven-Rawat: done

.github/workflows/remove-old-configuration-files.yml
Deven-Rawat: done

GeoffreyHayward (approved):

PR #2130, opened 2022-04-19
Dw 2887 refactor svg magic

Vlets:
SVG:
    - Social Media & Task List
        - Use “other” and upload own svg to use method 3.
    - NavigationBlock
        - Colors were hardcoded with svg magic. Frontend must add css
        - There are duplicate ftls? In folders Common and Nhsd-common. Problem is fixed in both.
    - Check if all places have the new method in place
    - Slowly phase out the old method


Vlets:
Please do not merge with master until missing CSS is added by a front-ender

PR #2131, opened 2022-04-20
[DW-2893] Remove boundary that may never match

GeoffreyHayward:


John-Milne (approved):
split on all the initial spaces* or spaces-

PR #2207, opened 2022-06-06
[DW-2957] Replace google verification file

lekeDevOps:


John-Milne (approved):
So removed a file and replaced it with another. 
Looks okay, assuming the new file is corect

PR #2218, opened 2022-06-16
[ADZ-1967] New Job to publish OAS specs from Proxygen PaaS endpoints

sophieclayton12-nhs:
[ADZ-1967: API spec - get OAS files from S3 bucket](https://nhsd-jira.digital.nhs.uk/browse/ADZ-1967)

### DIFF
--- a/cms/src/main/java/uk/nhs/digital/apispecs/ApiSpecificationPublicationService.java
+++ b/cms/src/main/java/uk/nhs/digital/apispecs/ApiSpecificationPublicationService.java
@@ -154,8 +154,11 @@ public class ApiSpecificationPublicationService {
     private void reportPublicationStatusFor(final int localSpecsCount, final int remoteSpecsCount, final SyncResults syncResults) {
 
         log.info(
-            "API Specifications found: in CMS: {}, in Apigee: {}, updated in Apigee and eligible to publish in CMS: {}, synced: {}, failed to sync: {}",
-            localSpecsCount, remoteSpecsCount, syncResults.eligible().size(), syncResults.published().size(), syncResults.failed().size()
+            "API Specifications found: in CMS: {}, in {}: {}, updated in {} and eligible to publish in CMS: {}, synced: {}, failed to sync: {}",
+            localSpecsCount, remoteSpecRepository.getClass().getSimpleName(),
+            remoteSpecsCount, remoteSpecRepository.getClass().getSimpleName(),
+            syncResults.eligible().size(), syncResults.published().size(),
+            syncResults.failed().size()
         );
 
         syncResults.published().forEach(specSyncResult ->
--- a/cms/src/main/java/uk/nhs/digital/apispecs/services/RemoteSpecService.java
+++ b/cms/src/main/java/uk/nhs/digital/apispecs/services/RemoteSpecService.java
@@ -1,7 +1,5 @@
-package uk.nhs.digital.apispecs.apigee;
+package uk.nhs.digital.apispecs.services;
 
-import static java.util.Collections.unmodifiableList;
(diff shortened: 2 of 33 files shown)


cms/src/main/java/uk/nhs/digital/apispecs/services/RemoteSpecService.java
```diff
@@ -16,21 +13,23 @@
 import java.util.List;
 import java.util.function.Supplier;
 
-public class ApigeeService implements OpenApiSpecificationRepository {
 
-    private static final Logger log = LoggerFactory.getLogger(ApigeeService.class);
+public abstract class RemoteSpecService implements OpenApiSpecificationRepository {
 
-    private static final String RESOURCE_NAMESPACE_APIGEE_MANAGEMENT_API = "apigeeManagementApi";
+    private static final Logger log = LoggerFactory.getLogger(RemoteSpecService.class);
```
lee-richardson-nhs, line 19:
```suggestion
    private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
```

If you use this then the log will show the extending class name (i.e. ProxygenService) rather than this abstract one.

pom.xml
```diff
@@ -903,7 +909,7 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-checkstyle-plugin</artifactId>
                 <version>${maven-chckstyle.version}</version>
-
+                <configuration><skip>true</skip></configuration>
```
lee-richardson-nhs, line 912: Should whip this out. I think this was us temporarily turning off checkstyle.

lee-richardson-nhs (changes requested):
Good work! Was a toughie to ones head around.

pom.xml
sophieclayton12-nhs: ah yes good spot! I've removed and resolved the checkstyle warning

cms/src/main/java/uk/nhs/digital/apispecs/services/RemoteSpecService.java
sophieclayton12-nhs: I tried doing this but the logs were still coming up as originating from parent class (RemoteSpecService). I even tried passing the instance of the Logger to the parent through the constructor but still no luck 🤔 
image

cms/src/main/java/uk/nhs/digital/apispecs/ApiSpecificationPublicationService.java
```diff
@@ -154,7 +154,7 @@ private Consumer<SpecificationSyncData> ifSuccessfulSoFar(final Consumer<Specifi
     private void reportPublicationStatusFor(final int localSpecsCount, final int remoteSpecsCount, final SyncResults syncResults) {
 
         log.info(
-            "API Specifications found: in CMS: {}, in Apigee: {}, updated in Apigee and eligible to publish in CMS: {}, synced: {}, failed to sync: {}",
+            "API Specifications found: in CMS: {}, in remote: {}, updated in remote and eligible to publish in CMS: {}, synced: {}, failed to sync: {}",
```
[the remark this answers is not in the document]
sophieclayton12-nhs: Good point, was doing some local testing earlier and wasn't clear from the logs. Have changed :)

cms/src/main/java/uk/nhs/digital/apispecs/services/RemoteSpecService.java
sophieclayton12-nhs: As a work around I have added the extending class name to the log text to make it clearer :)

image

PR #2232, opened 2022-06-28
[DW-2030] Events page broken

John-Milne:


site/components/src/main/java/uk/nhs/digital/common/components/FeedListComponent.java
```diff
@@ -180,5 +188,21 @@ public void doBeforeRender(HstRequest request,
...
+
+    @Override
+    public Filter createQueryFilter(HstRequest request, HstQuery hstQuery) throws FilterException {
+        return super.createQueryFilter(request, hstQuery);
+    }
+
+    @Override
+    public void applyAndFilters(HstQuery hstQuery, List<BaseFilter> filters) throws FilterException {
+        super.applyAndFilters(hstQuery, filters);
+    }
```
Deven-Rawat, line 206: Do we need to create above methods here?
getComponentParametersInfo
createQueryFilter
applyAndFilters

site/components/src/main/java/uk/nhs/digital/common/components/EventsComponent.java
```diff
@@ -31,6 +31,7 @@
 public class EventsComponent extends EssentialsEventsComponent {
 
     private static Logger log = LoggerFactory.getLogger(EventsComponent.class);
+    ComponentHelper componentHelper = new ComponentHelper();
```
Deven-Rawat, line 34: private?

site/components/src/main/java/uk/nhs/digital/common/components/EventsComponent.java
```diff
@@ -114,4 +120,20 @@ protected int getPageSize(final HstRequest request, final EssentialsPageable par
         return NumberUtils.isCreatable(compononentPageSize)
             ? Integer.parseInt(compononentPageSize) : super.getPageSize(request, paramInfo);
     }
+
```
Deven-Rawat, line 123: Do we need to create above methods here?
getComponentParametersInfo
createQueryFilter
applyAndFilters

site/components/src/main/java/uk/nhs/digital/common/components/FeedListComponent.java
John-Milne: Yes, as the super methods have protected access modifiers and thus are not accessible outside of their class or sub-class

site/components/src/main/java/uk/nhs/digital/common/components/EventsComponent.java
John-Milne: Yes, as the super methods have protected access modifiers and thus are not accessible outside of their class or sub-class

PR #2251, opened 2022-07-12
[DW-3070] Added Image RSS tag

Deven-Rawat:


site/components/src/main/java/uk/nhs/digital/common/feed/RssModifier.java
```diff
@@ -261,11 +256,32 @@ public void modifyEntry(final HstRequestContext context, final Item entry, final
     @NotNull
     private Element getImageElement(String imageUrl) {
         final Element element
-            = new Element("thumbnail", "media", "http://search.yahoo.com/mrss/");
+            = new Element("content", "media", "http://search.yahoo.com/mrss/");
```
racanosman, line 259: Suggest you put this URL in a property file or resource bundle rather than hardcode it in this class.

site/components/src/main/java/uk/nhs/digital/common/feed/RssModifier.java
```diff
@@ -261,11 +256,32 @@ public void modifyEntry(final HstRequestContext context, final Item entry, final
...
+            contentType = connection.getContentType();
+            element.setAttribute("type", contentType.substring(0, contentType.indexOf(";")));
+            BufferedImage bufferedImage = ImageIO.read(url);
+            if (bufferedImage != null) {
+                element.setAttribute("height", String.valueOf(bufferedImage.getHeight()));
+                element.setAttribute("width", String.valueOf(bufferedImage.getWidth()));
+            }
+            element.setAttribute("expression", "full");
+        } catch (IOException e) {
+            e.printStackTrace();
```
racanosman, line 281: Use log error instead of just printing the stack trace

site/components/src/main/java/uk/nhs/digital/common/feed/RssModifier.java
Deven-Rawat: Can't be done atm

site/components/src/main/java/uk/nhs/digital/common/feed/RssModifier.java
Deven-Rawat: done

PR #2254, opened 2022-07-18
[DW-2989] Commit social media doc type permanently

lekeDevOps:


John-Milne (approved):
Looks fine and already tested so approved.

PR #2265, opened 2022-07-25
[DW-2915] Fixed deprecated code

Deven-Rawat:


John-Milne (approved):
Issue in PlainTextStatisticHttpRequestInterceptor causing test to fail, this is corrected in DW-2916 so not to worry here

PR #2296, opened 2022-08-05
[ADZ-543] ADZ-543 adding REST endpoint to get OAS JSON data

harryhmc:


sophieclayton12-nhs (commented):
Looks great to me, cool to see the bloomreach rest endpoint implementation! 👏 

I may have missed something, but I thought we were going to have both the download button at the top of the spec and a link in the try this api swagger bit to download the spec? Was there a decision to just do the button in the end? Just wanted to check 🙂 

I'm not sure about the `_target="blank"` thing you mentioned, I guess it makes sense for it to be the same window as the user isn't navigating away from the site?

harryhmc:
> Looks great to me, cool to see the bloomreach rest endpoint implementation! :clap:
> 
> I may have missed something, but I thought we were going to have both the download button at the top of the spec and a link in the try this api swagger bit to download the spec? Was there a decision to just do the button in the end? Just wanted to check :slightly_smiling_face:

this can be implemented, but i think it'd be good to have a design aspect for this first and maybe a separate ticket. it involves some CSS changes targeting specific containers, as well as a simple change in `api-try-it-now.ftl`
> I'm not sure about the `_target="blank"` thing you mentioned, I guess it makes sense for it to be the same window as the user isn't navigating away from the site?

i think i'd prefer a separate tab, as there's no way to get back from the result without using the browser's back button. similarly, the resulting webpage isn't actually a page on our website so much as just raw JSON. opening it in the same tab feels like interrupting the user flow to me?

sophieclayton12-nhs (approved):
Ahh makes sense, separate ticket could be good to progress this too and the button alone may be enough anyways.

I agree with your separate tab points, it was hard to visualise the flow in my head but now think separate tab is the way to go :D

lee-richardson-nhs (approved):

PR #2298, opened 2022-08-08
[DW-2879] Fixed broken sticky nav link

LEJA3:


John-Milne (approved):
Looks fine, removing the double quotes from around the section heading as Freemarker adds them

PR #2336, opened 2022-09-12
[DW-2978] Removed sticky positioning from chapter nav

LEJA3:


John-Milne (approved):
All looks to be in order

PR #2340, opened 2022-09-14
[DW-3081] Updated icon sizes to account for padding

LEJA3:


John-Milne (approved):
All looks to be in order

PR #2344, opened 2022-09-15
[DW-3155] Manage date-time formats for automatically loaded content

ianp-mani:
Update to the processing for date-time formats in the automatic report creation work. It now recognises full or partial dates and adds those appropriately rather than always defaulting to midnight


cms/src/main/java/uk/nhs/digital/arc/util/DateHelper.java
```diff
@@ -7,27 +7,50 @@
...
+                return fullLengthDate(firstDate);
             }
         }
 
-        return "";
+        return null;
+    }
+
+    /**
+     * We get a wel formatted date but it may be opartial, so we'll pad it out with default values if required
```
GeoffreyHayward, line 38: Typo in well.

GeoffreyHayward (approved):

PR #2347, opened 2022-09-20
[DW-3074] Set up export service for Highcharts

LEJA3:


site/components/src/main/java/uk/nhs/digital/website/rest/DataVizResource.java
```diff
@@ -0,0 +1,71 @@
...
+            if (!iterator.hasNext()) {
+                return null;
+            }
+            HippoBean document = iterator.nextHippoBean();
+            List<ChartSection> sections = document.getChildBeans(ChartSection.class);
+
+            ChartSection section = sections.stream()
+                .filter(s -> s.getUniqueId().equals(chartId))
+                .findFirst()
+                .orElse(null);
```
GeoffreyHayward, line 60: This stream is okay, as long as you know the list size will not get too long. This is because all ChartSection objects get loaded into memory before applying the filter.

GeoffreyHayward (approved):
Approved, but I left one comment to consider.

PR #2407, opened 2022-11-23
[DW-3238] Updated ipfilter

Deven-Rawat:


repository-data/application/src/main/resources/hcm-config/configuration/modules/ipfilter.yaml
```diff
@@ -15,13 +15,25 @@ definitions:
...
+        - cms-cs2.nhsd.io
+        - cms-development.nhsd.io
+        - cms-training2.nhsd.io
+        - cms-uat2.nhsd.io
+        - cs2.nhsd.io
+        - development.nhsd.io
+        - training2.nhsd.io
+        - uat2.nhsd.io
+        - cms-test.nhsd.io
+        - test.nhsd.io
```
GeoffreyHayward, line 36: For readability, it would be better to group the CMS address with the corresponding SITE address. For example putting these together:

- cms-uat2.nhsd.io
- uat2.nhsd.io

Like you have with the test env.

GeoffreyHayward (commented):
Looks great. 

From production, please remove hostname `digital.nhs.uk` as the site should not be filtered.

Also, please add a block that is the same a production, but named `production-candidate` with the single hostname `cms.nhsd.io`.

GeoffreyHayward (approved):
Nice work.

PR #2464, opened 2023-02-13
[DW-3137] Set up series publication search

lekeDevOps:


site/components/src/main/java/uk/nhs/digital/ps/beans/Series.java
```diff
@@ -255,4 +257,13 @@ protected <T extends HippoBean> List<T> getRelatedDocuments(
         return toList(query.execute().getHippoBeans());
     }
 
+    // because there is no publicationsystem:NominalDate variable for series we need to use either
```
lee-richardson-nhs, line 260: This would be more visible if it were a JavaDoc so it would appear in the doc pop ups. And then you'd need to specify which field is actually being used.

lekeDevOps: thanks @lee-richardson-nhs I will ensure this is done going forward. :)

lee-richardson-nhs (approved):

PR #2496, opened 2023-04-14
[ADZ-2719] Thrown Execption for empty param list

uzair-haroon:


site/components/src/main/java/uk/nhs/digital/common/components/apispecification/swagger/request/examplerenderer/CodegenParameterExampleHtmlRenderer.java
```diff
@@ -102,7 +102,10 @@ private String htmlFrom(final Collection<ParamExample> complexExamplesFromParamS
                         return examplesHtml;
                     })
                     .reduce(StringBuilder::append)
-                    .get()
+                    .orElseThrow(() -> new NullPointerException(
```
lee-richardson-nhs, line 105: If it were a null pointer it would have thrown before this point, on trying to access the object. `IllegalArgumentException` would be better.

lee-richardson-nhs (approved):

PR #2526, opened 2023-06-14
AMP-1653-service-catalogue

clpo:
[Ticket](https://nhsd-jira.digital.nhs.uk/browse/AMP-1653)

Service catalogue just like the existing API Catalogue.

Some renames of files and package to more generic 'Catalogue' given that they aren't coupled to API spec links now.


site/components/src/main/java/uk/nhs/digital/common/components/catalogue/ApiCatalogueComponent.java
```diff
@@ -0,0 +1,84 @@
+package uk.nhs.digital.common.components.catalogue;
```
clpo, line 1: Some files are showing as new here but really they have just a few changes but have had their package renamed.

lee-richardson-nhs (approved):

GeoffreyHayward (approved):
Looks good.

site/components/src/main/java/uk/nhs/digital/common/components/catalogue/ApiCatalogueComponent.java
clpo: line 19 has the mapping document path now, there a new request parameter line 48. some other stuff that was in this class is now in CatalogueComponent for inheritance.

PR #2528, opened 2023-06-20
[DW-3392] Tidy of makefiles and dev docs

GeoffreyHayward:


GeoffreyHayward:
Changes

- Promotes body of make serve.enhance to make serve and drops __ latter. __ 
- Drops essentials/target/essentials.war as didn't seem to work.
- Drops prep.s3-perf-test-data as never used. 
- Drops update-dependencie as it didn't work.
- Update's main Readme.md to keep instructions inline.
- Removes unused env args in the makefile.
- Drops aws-sudo as never used.
- Drops test.wip.standalone as never used. 
- Drops vendor as never used.
- Removes generate-s3-perf-test-docs from the main pom.xml as it's corresponding make command has been dropped. 
- Drops GenerateTestDocsForTestingS3Concurrent.groovy as it's corresponding make command has been dropped.

Vlets (approved):

PR #2559, opened 2023-07-25
AMP-1861-proxygen-auth-issue

clpo:
Getting an access token for use in a call to get proxygen specs is failing. Seems like this is due to a missing KID header.

Have already updated the config properties for UAT and CS environments in testing. Need to update production config in order for this to work upon deployment.

### DIFF
--- a/acceptance-tests/pom.xml
+++ b/acceptance-tests/pom.xml
@@ -356,7 +356,7 @@
                                     <platform>linux</platform>
                                     <bit>64</bit>
                                     <version>${chromedriver.version}</version>
-                                    <url>https://edgedl.me.gvt1.com/edgedl/chrome/chrome-for-testing/${chromedriver.version}/linux64/chromedriver_linux64.zip</url>
+                                    <url>https://edgedl.me.gvt1.com/edgedl/chrome/chrome-for-testing/${chromedriver.version}/linux64/chromedriver-linux64.zip</url>
                                 </driver>
                                 <driver>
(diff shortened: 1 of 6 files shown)


acceptance-tests/pom.xml
```diff
@@ -356,14 +356,14 @@
                                     <platform>linux</platform>
                                     <bit>64</bit>
                                     <version>${chromedriver.version}</version>
-                                    <url>https://chromedriver.storage.googleapis.com/${chromedriver.version}/chromedriver_linux64.zip</url>
+                                    <url>https://edgedl.me.gvt1.com/edgedl/chrome/chrome-for-testing/${chromedriver.version}/linux64/chromedriver-linux64.zip</url>
```
clpo, line 359: The chromedriver.storage site does not have the required version of chromedriver. https://chromedriver.chromium.org/downloads suggests to use the other source for 115.0.5790.102

pom.xml
```diff
@@ -162,7 +162,7 @@
         Consult https://chromedriver.chromium.org/downloads
         to find the right ChromeDriver version to use with the current browser version,
         -->
-        <chromedriver.version>113.0.5672.63</chromedriver.version>
+        <chromedriver.version>115.0.5790.102</chromedriver.version>
```
clpo, line 165: The pr environment is using the equivalent version of chrome so this chromedriver version is now required.

GeoffreyHayward (approved):

PR #2595, opened 2023-10-03
Amp 1380 multiple filters in category

clpo:
Multiple select of filters for Service/API catalogues.
Filters to apply to the entries with OR within the same section and AND with other sections, similar to how it exists in many filter systems around the web. Previously filters within the same section would be mutually exclusive to each other applying the filtering as AND always, unless an entry happened to have both filter tags.
This also must apply to the Filter nav itself in that filters within the same category as just selected will remain to give the user opportunity to select some other filters before they select another sections' filter.

[Ticket](https://nhsd-jira.digital.nhs.uk/browse/AMP-1380)


site/components/src/main/java/uk/nhs/digital/common/components/catalogue/filters/FiltersAndLinks.java
```diff
@@ -0,0 +1,117 @@
...
+                filteredLinks.set(filteredLinksForCategory);
+                filtersForCategoryFromLinks.addAll(filtersNotInCollection(filteredLinksForCategory, filtersForCategoryFromLinks));
+                addToFilters(filtersForCategoryFromLinks);
+                Set<String> allFilteredLinksKeys = filteredLinks.get().stream().flatMap(link -> link.allTaxonomyKeysOfReferencedDoc().stream()).collect(Collectors.toSet());
+                removeFromFilters(filters.parallelStream().filter(key -> !collectionsContainKey(category, allFilteredLinksKeys, key)).collect(Collectors.toSet()));
+            });
+        } else {
+            addToFilters(allKeysFromLinks(links));
+        }
+    }
```
GeoffreyHayward, line 50: As this is such a big chunk of code, please can you add some comments?

GeoffreyHayward (changes requested):
It seems okay, but are you happy to talk me through the changes on a call? That way I could provide better feedback.

site/components/src/main/java/uk/nhs/digital/common/components/catalogue/filters/FiltersAndLinks.java
clpo: yeah sure, it is quite a wall of code to be fair

GeoffreyHayward (approved):

PR #2600, opened 2023-10-24
[AMP-1990] Set referrer policy for catalogues

clpo:
We have a link to a Qualtrics survey that is set up to consume the referrer header of the incoming request. The link lives in the content of an update which exists on a number of pages.
Issue was that the referrer header was always empty upon arriving on the survey page which meant that this was not being consumed by the survey.

Change is to the configuration of the Html processor to stop it adding a 'noreferrer' flag for the link.

### DIFF
--- a/repository-data/application/src/main/resources/hcm-config/namespaces/website/update.yaml
+++ b/repository-data/application/src/main/resources/hcm-config/namespaces/website/update.yaml
@@ -34,10 +34,10 @@ definitions:
               ckeditor.config.overlayed.json: '{linkShowAdvancedTab: true, extraPlugins:''wordcount'',
                 disableNativeSpellChecker: false}'
               htmlprocessor.id: richtext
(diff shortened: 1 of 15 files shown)


repository-data/application/src/main/resources/hcm-config/namespaces/website/update.yaml
```diff
@@ -33,6 +33,7 @@ definitions:
                 items: [''Link'',''PickInternalLink'',''Source''] }] }'
               ckeditor.config.overlayed.json: '{linkShowAdvancedTab: true, extraPlugins:''wordcount'',
                 disableNativeSpellChecker: false}'
+              htmlprocessor.id: no-filter
```
clpo, line 36: There is a more specific option than this but it didn't seem to work for me.

GeoffreyHayward (approved):

PR #2635, opened 2024-02-06
[AMP-2144] Catalogues - Show only first few filters and default sections expanded

clpo:
Create new configurable fields in the taxonomy mapping for catalogues to control:
- Default expanded sections
- Hide subsections
- Number of subsections to show

Update YAML with:
defaultExpanded: true
hideChildren: true
amountChildrenToShow: 2

Implemented logic to make use of these fields.
There is a 'show more' link that will expand the section to show all subsections.

**With subsections hidden**
image

**Subsections shown**
image


acceptance-tests/src/test/resources/features/site/apiCatalogue.feature
```diff
@@ -21,8 +21,6 @@ Feature: API Catalogue in Developer hub
         And I should see elements with attributes:
             | text         | class                                                                             |
             | Inpatient    | nhsd-a-tag nhsd-a-tag--bg-light-grey nhsd-!t-margin-top-3 nhsd-!t-margin-bottom-1 |
```
clpo, line 23: I removed demographics from the catalogue filters which is why I removed this case from the test.

repository-data/development/src/main/resources/hcm-content/content/documents/corporate-website/api-catalogue.yaml
```diff
@@ -694,17 +694,15 @@
...
-          - data-and-information
-          - inpatients
-          - national-health-service
-          - hospital
-          - secondary-care
-          - demographics
+          - security
+          - patient-communication
+          - reference-data
+          - tests-and-diagnostics
```
GeoffreyHayward, line 700: As this file is in the content directory, you will also need to make all of the changes in this file in the console too.

repository-data/local/src/main/resources/hcm-content/content/taxonomies/publication_taxonomy.yaml
```diff
@@ -332,6 +332,14 @@
...
       hippotaxonomy:key: outcomes-and-quality-indicators
       jcr:primaryType: hippotaxonomy:category
+    /patient-communication:
+      /hippotaxonomy:categoryinfos:
+        /en:
+          hippotaxonomy:name: Patient communication
+          jcr:primaryType: hippotaxonomy:categoryinfo
+        jcr:primaryType: hippotaxonomy:categoryinfos
+      hippotaxonomy:key: patient-communication
+      jcr:primaryType: hippotaxonomy:category
```
GeoffreyHayward, line 342: Same with this file, you will need to make these changes in the console too.

GeoffreyHayward (approved):

repository-data/development/src/main/resources/hcm-content/content/documents/corporate-website/api-catalogue.yaml
vinodkumarnarayanan1: Thanks for the feedback. As I checked, this two yaml file changes in the content directory are used only for the integration test run, not used at any functional level. They run off with the the local data.

repository-data/local/src/main/resources/hcm-content/content/taxonomies/publication_taxonomy.yaml
vinodkumarnarayanan1: same comment as above, it is used only for the integration test to run against local data.

PR #2665, opened 2024-04-29
[AMP-2732] Fix for handling broken special character '&' in catalogue…

vinodkumarnarayanan1:
This fix for handling the broken special character - '&' after implementing yellow highlight match string in search pages in both Service and API catalogues.

Before Fix:
Screenshot from 2024-04-29 16-14-30

After Fix:
image

### DIFF
--- a/repository-data/webfiles/src/main/resources/site/apicatalogue/apicatalogue.js
+++ b/repository-data/webfiles/src/main/resources/site/apicatalogue/apicatalogue.js
@@ -67,9 +67,10 @@ function findHTMLElementData(result, element, htmlElement) {
 function replaceHTMLContent(result, element, text, searchTerm, htmlElement) {
     const highlightContent1 = text.replaceAll('<span class="filter-tag-yellow-highlight">', '');
     const highlightContent2 = highlightContent1.replaceAll('</span>', '');
+    const highlightContent3 = highlightContent2.replaceAll('&amp;', '&');
     let highlightContentFinal = '';
     if (searchTerm.length !== 0) {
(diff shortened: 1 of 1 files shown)


repository-data/webfiles/src/main/resources/site/apicatalogue/apicatalogue.js
```diff
@@ -67,19 +67,20 @@ function findHTMLElementData(result, element, htmlElement) {
 function replaceHTMLContent(result, element, text, searchTerm, htmlElement) {
     const highlightContent1 = text.replaceAll('<span class="filter-tag-yellow-highlight">', '');
     const highlightContent2 = highlightContent1.replaceAll('</span>', '');
+    const highlightContent3 = highlightContent2.replaceAll('&amp;', '&');
```
GeoffreyHayward, line 70: These three `replaceAll` statements can be combined to save on intermediate variables. 
```
const highlightContent = text
    .replaceAll('<span class="filter-tag-yellow-highlight">', '')
    .replaceAll('</span>', '')
    .replaceAll('&amp;', '&');
```
By chaining these calls together you improve clarity.

GeoffreyHayward (approved):
Approved, but please consider updating it inline with my comment.

repository-data/webfiles/src/main/resources/site/apicatalogue/apicatalogue.js
vinodkumarnarayanan1: This is amended as above. Thank you for your comments and approval @GeoffreyHayward.

GeoffreyHayward: No problem.

PR #2702, opened 2024-10-18
[AMP-3121] Api Catalogue junit tests

vinodkumarnarayanan1:


site/components/src/main/java/uk/nhs/digital/common/components/catalogue/ApiCatalogueHubComponent.java
```diff
@@ -64,15 +64,16 @@ public void doBeforeRender(final HstRequest request, final HstResponse response)
...
         long endTime = System.currentTimeMillis();
         long duration = endTime - startTime;
-        log.info("End of method: doBeforeRender in ApiCatalogueHubComponent  at " + endTime + " ms. Duration: " + duration + " ms");
+        log.info("End of method: doBeforeRender in ApiCatalogueHubComponent  at {} ms. Duration: {} ms", endTime, duration);
     }
 
     @Override
-    protected <T extends EssentialsListComponentInfo> Pageable<HippoBean> doFacetedSearch(HstRequest request, T paramInfo, HippoBean scope) {
-
+    public Pageable<HippoBean> doFacetedSearch(HstRequest request, EssentialsListComponentInfo paramInfo, HippoBean scope) {
```
GeoffreyHayward, line 71: What's the reason for this change? Leaving it as it was, `ApiCatalogueHubComponentTest` will continue to compile because it is already package and subclass visible. 

Moreover, we lose flexibility if we need to subclasses or extend its behaviour in the future. The method no longer accepts a parameter of any subclass of `EssentialsListComponentInfo`. Instead, it now strictly expects an `EssentialsListComponentInfo`.

GeoffreyHayward: Also, I am not convinced this modified method (without the generic type parameter) would be considered an override anymore.

GeoffreyHayward (changes requested):

site/components/src/main/java/uk/nhs/digital/common/components/catalogue/ApiCatalogueHubComponent.java
```diff
@@ -18,44 +18,44 @@
...
 public class ApiCatalogueHubComponent extends EssentialsListComponent {
 
     private static final Logger log = LoggerFactory.getLogger(ApiCatalogueHubComponent.class);
 
+    private ApiCatalogueFilterManager apiCatalogueFilterManager;
+
+    public void setApiCatalogueFilterManager(ApiCatalogueFilterManager apiCatalogueFilterManager) {
+        this.apiCatalogueFilterManager = apiCatalogueFilterManager;
+    }
+
```
GeoffreyHayward, line 36: ### Part 1/2 (unintended shared state)

_The original comment was lost :(, I try again_ 

**Once this is set, it's set!** To avoid side effects, change this to have two constructors. 

```
ApiCatalogueHubComponent() {
  /* Used by the framework */
}

ApiCatalogueHubComponent(ApiCatalogueFilterManager apiCatalogueFilterManager) {
  this.testableApiCatalogueFilterManager = apiCatalogueFilterManager;
}
```

site/components/src/main/java/uk/nhs/digital/common/components/catalogue/ApiCatalogueHubComponent.java
```diff
@@ -18,44 +18,44 @@
...
         long startTime = System.currentTimeMillis();
         log.debug("Start Time:" + startTime);
 
         super.doBeforeRender(request, response);
 
-        ApiCatalogueFilterManager apiCatalogueFilterManager = new ApiCatalogueFilterManager();
+        // This is to create the object at first time when the object is not created through junit.
+        if (apiCatalogueFilterManager == null) {
+            apiCatalogueFilterManager = new ApiCatalogueFilterManager();
+        }
```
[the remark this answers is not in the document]
vinodkumarnarayanan1: Changed the code as suggested above.

site/components/src/main/java/uk/nhs/digital/common/components/catalogue/ApiCatalogueHubComponent.java
```diff
@@ -18,44 +18,44 @@
...
-                    entry -> entry
-                ));
+            Map<String, Section> sectionMap = section.getEntries().stream().collect(Collectors.toMap(Section::getTaxonomyKey, entry -> entry));
             request.setAttribute("apiStatusEntries", sectionMap);
         });
 
-        List<Section> nonStatusSections = sections.stream()
-            .filter(section -> !"Status".equals(section.getDisplayName()))
-            .collect(Collectors.toList());
+        List<Section> nonStatusSections = sections.stream().filter(section -> !"Status".equals(section.getDisplayName())).collect(Collectors.toList());
```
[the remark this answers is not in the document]
vinodkumarnarayanan1: Changed the code format as like master. please check now.

site/components/src/main/java/uk/nhs/digital/common/components/catalogue/ApiCatalogueHubComponent.java
vinodkumarnarayanan1: Reverted the code as like master.

site/components/src/main/java/uk/nhs/digital/common/components/catalogue/ApiCatalogueHubComponent.java
vinodkumarnarayanan1: As discussed, constructor method doesn't helped here, getting some errors. Since we changed the object creation as Part 2/2 approach and we are not using threading here, unintended shared stared issue will not occur here.

GeoffreyHayward (approved):

GeoffreyHayward:
Please rebase to 1 commit before merging.

PR #2703, opened 2024-10-22
[AMP-3153] Fixed header height adjustment fix in api spec page

vinodkumarnarayanan1:
In API specification page, made the changes to display the section header('Supplier Statuses' in the below example) in top of the main body when user clicks its section links from another section main body.  

It was not worked for the rapidoc generated api specs, it is fixed in the story in rapidoc css.

Before Fix:
image

After Fix:
image

Anchor links div contents was not displaying properly when we refresh few page with anchor links, page was scrolling up and the section header was hiding top.
I have fixed this issue by adjusting the windows scroll offset position. 

Before Fix: (During Page refresh)
image

After Fix: (During Page refresh)
image


vinodkumarnarayanan1:
@GeoffreyHayward  - Rapidoc js is a minified js file, not able to see the exact  code difference in git, so please see the code difference here for your review.
image

PR #2709, opened 2024-12-03
[WSP-270] Workflows change to take new relic properties

sankar-shunmuga:
Following workflows change to take new relic properties 
1. .github/workflows/slack-slash-deploy.yml
2. .github/workflows/deploy-to-production-stage-2.yml

Since the trial version of new relic has limited monthly data ingestion, the idea is to apply the properties file to production environment only. Slack-slash-deploy.yml will be used only for test purposes on lower environments and it will be reverted after completion of PR review & testing.


sankar-shunmuga:
This pull request is not required now.

PR #2731, opened 2025-02-20
[WSP-353] Adds 'Should run anyway' mode

GeoffreyHayward:


sankar-shunmuga (approved):
thanks @GeoffreyHayward  it looks good

PR #2735, opened 2025-03-10
WSP-291-Headings not capture

FranciscoRuizHEE:
A validation was added to avoid the html section to be created if the "section.heading" is empty


sankar-shunmuga (approved):
thanks @FranciscoRuizHEE , approved

PR #2738, opened 2025-03-19
[WSP-396] Publication job fix

GeoffreyHayward:
Before the fix
- It would not retry to update the document if, for any reason, the job failed; now, it will try again the time the job runs.
- The whole job would halt if any document had an unmerged edit; now, it will continue to the next document.

**Now, safely, documents that have unmerged edits will continue to fail until a content editor has approved the changes. But this will no longer fail the whole job.** Once the document has been merged or reverted, the next run will pick it up correctly. 

As well as the fix, I have added:
- For future maintenance, comments that explain the code.
- Unit tests

### DIFF
--- a/cms/src/main/java/uk/nhs/digital/ps/PublicationAccessibilityUpdateJob.java
+++ b/cms/src/main/java/uk/nhs/digital/ps/PublicationAccessibilityUpdateJob.java
@@ -12,7 +12,6 @@ import org.onehippo.repository.scheduling.RepositoryJobExecutionContext;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.time.LocalDate;
 import java.time.LocalDateTime;
 import java.util.Calendar;
 import javax.jcr.Node;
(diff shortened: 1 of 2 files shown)


cms/src/main/java/uk/nhs/digital/ps/PublicationAccessibilityUpdateJob.java
```diff
@@ -21,67 +20,127 @@
...
+                        JcrUtils.ensureIsCheckedOut(documentNode);
+                        LOGGER.debug("Marking publication as publicly accessible at: {}", pubNode.getPath());
+
+                        // Update properties to reflect the new publicly accessible state
+                        documentNode.setProperty("publicationsystem:PubliclyAccessible", true);
+                        documentNode.setProperty(PublicationSystemConstants.PROPERTY_EARLY_ACCESS_KEY, "");
+                        documentNode.setProperty("hippostdpubwf:lastModificationDate", Calendar.getInstance());
+
+                        // Save changes to the repository
+                        session.save();
```
davidbrrr, line 104: I don't think the session needs to be saved because the documentManager will do that, but please test this out if you remove it.

davidbrrr (approved):

PR #2739, opened 2025-03-20
[WSP-400]-Deploy to Prod Stage 1 & 2 - cron job, clocks move

FranciscoRuizHEE:
Change to BST time in Git actions workflow Prod Stage 1 and 2


GeoffreyHayward (approved):
@FranciscoRuizHEE that's great. But can you also update deploy-to-production-stage-1.yml?

PR #2742, opened 2025-03-26
WSP-385-Two same( i.e. H1) level headings in a page

FranciscoRuizHEE:
Some h1 tags had a incorrect heading level, they needed to be h4 or h5


GeoffreyHayward (approved):
Looks good.

PR #2744, opened 2025-04-09
Wsp 324 fix headings

FranciscoRuizHEE:
Create some Cucumber test to ensure headers accessibility standards


sankar-shunmuga (approved):
thanks @FranciscoRuizHEE , approved

PR #2757, opened 2025-07-22
[WSP-634] Adds Windows Dev Set Steps

GeoffreyHayward:


docs/what-if/first-time-set-up-windows.md
```diff
@@ -0,0 +1,266 @@
...
+```bash
+# Note you can skip these command as you
+# should have already run `apt update` in step 2.
+sudo apt update
+```
+
+To install update apt.
+
+```bash
+sudp apt install maven
```
sankar-shunmuga, line 114: sudo apt install maven

PR #2814, opened 2026-02-25
[WSP-1029] Null check SearchEssentialsFacetsComponent

sankar-shunmuga:


FranciscoRuizHEE (approved):

GeoffreyHayward:
The change to `site/components/src/main/java/uk/nhs/digital/common/components/SearchEssentialsFacetsComponent.java` and its new Unit Tests address the primary issue in the ticket. However, during testing, a new issue was found due to a misconfiguration in the `/pages/main/left`, still looking for `facets` instead of `faceted-search`. Hence the additional YAML config changes.

PR #2824, opened 2026-04-09
[WSP-1083] Unable to deploy from master to UAT - bugfix

sankar-shunmuga:


GeoffreyHayward (approved):
The issue was the raw output flag `jq -r`, instead of `jq`. Raw output is not a JSON string.

PR #2826, opened 2026-04-17
[WSP-922] Impersonate session in daemons classes

FranciscoRuizHEE:


GeoffreyHayward (commented):
Thanks for the work on this – the direction looks good and aligns with the recommendation to avoid long-lived daemon sessions.

### Key Feedback

The main objective here is to avoid using a cached session from `initialize(...)` and instead ensure that each event is handled with a fresh impersonated session that is properly closed afterwards. From reviewing the changes, the implementation is broadly correct and moving in the right direction.

One thing I would suggest tightening is the reasoning behind the change. The primary issue is not thread safety, but rather that long-lived sessions can become stale, invalid, or logged out. If that happens, the listener may silently stop functioning until the application is restarted. Using `Session.impersonate()` per event ensures a valid session lifecycle for each invocation and avoids this failure mode.

### Unit Tests

The current PR is missing unit tests, which are important here because we are touching session lifecycle behaviour.

We should aim to cover the following scenarios:

* Verify that `Session.impersonate()` is called when an event is handled
* Verify that the impersonated session is used for processing (and not the original daemon session)
* Verify that the impersonated session is always logged out after successful processing
* Verify that the impersonated session is logged out even when an exception occurs
* Verify that existing business behaviour of the listener remains unchanged (happy path)

This gives us confidence that the change is both safe and correctly implemented.

### PreReleaseKeyListener

I noticed that `PreReleaseKeyListener` has not yet been updated in this PR, but it follows the same long-lived session pattern and is part of the original scope.

To help with this, I’ve raised a separate PR for you to merge into your branch, and then we can merge as part of the final work you are doing. I have added:

* An implementation of session impersonation for `PreReleaseKeyListener`
* Unit tests demonstrating the expected approach

PR: [https://github.com/NHS-digital-website/hippo/pull/2831](https://github.com/NHS-digital-website/hippo/pull/2831)

This should serve as both a reference for how to apply the same pattern and testing approach to the other listener classes in this ticket, as well as the actual tests for `PreReleaseKeyListener`.

### PublicationPostListener

This class appears to have been inactive (listener registration was commented out), but you've removed the commented-outness. Please can you verify this was intentional, as this will, of course, affect the behaviour? 

### Summary

* Implementation direction is correct
* Reasoning should focus on session lifecycle (not thread safety)
* Unit tests are required to validate impersonation and session cleanup
* `PreReleaseKeyListener` still needs to be addressed (work provided in the linked PR)

Let me know if you want me to help align the remaining listeners to the same pattern.

FranciscoRuizHEE:
@GeoffreyHayward I have checked the changes, and the PreReleaseKeyListener changes were made in the ProcessSearch file, where all the logic is and the session is used. It needs to be in one or the other. For the unit test should they be needed for each class?

cms/src/main/java/uk/nhs/digital/common/listeners/PublicationPostListener.java
```diff
@@ -69,20 +72,24 @@ private void addTierLevelToNode(HippoWorkflowEvent event) {
                     }
                 }
             }
-            //session.save();
+            session.save();
```
GeoffreyHayward: The code modifies nodes through `newSession`, then calls `session.save()` on the original daemon session. Those transient changes belong to `newSession`, so they will not be persisted before `newSession.logout()` discards them. This should be `newSession.save()`.

GeoffreyHayward (changes requested):
Please see comment

GeoffreyHayward (approved):

PR #2829, opened 2026-04-21
[WSP-1036] Search Component Null value fix

sankar-shunmuga:


site/components/src/main/java/uk/nhs/digital/common/components/SearchComponent.java
```diff
@@ -17,70 +18,226 @@
...
+
+    private static final String PAGEABLE_MODEL = "pageable";
+    private static final String TOTAL_RESULTS_ATTR = "totalResults";
+    private static final String SORT_ATTR = "sort";
+
+    /**
+     * UUID-like / identifier-like values are not normal free-text search terms and can
+     * cause issues in faceted navigation when passed through as a free query.
+     */
+    private static final Pattern UUID_PATTERN = Pattern.compile(
```
GeoffreyHayward: Using the JDK parser is usually simpler and more reliable:

```
public static boolean isValidUuid(String value) {
    try {
        UUID.fromString(value);
        return true;
    } catch (IllegalArgumentException e) {
        return false;
    }
}
```

This would avoids maintaining a regex and ensures the format matches what Java itself accepts.

GeoffreyHayward (approved):
Looks good.

site/components/src/main/java/uk/nhs/digital/common/components/SearchComponent.java
sankar-shunmuga: @GeoffreyHayward as suggested, I removed the regex-based pattern check and replaced it with your recommended UUID checker. Thanks.

site/components/src/main/java/uk/nhs/digital/common/components/SearchComponent.java
```diff
@@ -22,65 +21,148 @@ public class SearchComponent extends EssentialsListComponent {
...
+    private static final String PAGEABLE_MODEL = "pageable";
+
     @Override
     public void doBeforeRender(HstRequest request, HstResponse response) {
         long startTime = System.currentTimeMillis();
-        LOGGER.debug("SearchComponent2 - Start Time:" + startTime);
+        LOGGER.debug("SearchComponent - Start Time: {}", startTime);
+
+        final String query = SearchQuerySanitizer.getSanitizedSearchQuery(request, this);
+        request.setAttribute(QUERY_PARAM, query);
```
GeoffreyHayward, line 36: Some possible redundancy where as super() does this too.

GeoffreyHayward (approved):
Looks good

PR #2852, opened 2026-06-09
[WSP-1110] Fix for menu tabbing

GeoffreyHayward:
PR ready for review and testing. 

I have implemented a small runtime enhancement for the responsive NHS Digital global header menu keyboard behaviour.

  Summary of change:

  - Added a new global-header-focus-order.js module to correct keyboard focus order when the responsive global header menu is open.
  - Added a separate global-header-escape-close.js module so pressing Escape closes the open menu via the existing design-system menu button handler and returns focus to the menu
    button.

  - Wired both modules into the NHS Digital and NDRS frontend script entry points.
  - Also lint-formatted ndrs-frontend-scripts.js, as the Maven build lints all changed JS files against origin/master.

  Behaviour covered:

  - Menu content is removed from tab order while the responsive menu is closed.
  - When the menu is opened with keyboard, focus remains on the Menu button until the user tabs forward.
  - Tab from the Menu button moves into the menu links.
  - Shift+Tab from the first menu item moves back to the Menu button.
  - Shift+Tab from the Menu button closes the menu and allows focus to continue back up the page.
  - Tab from the last menu item closes the menu and moves focus to the next focusable element after the header, for example the breadcrumb link, rather than jumping back to the
    Search button.

  - Activating Close menu closes the menu and returns focus to the Menu button.
  - Escape closes the menu only when it is actually open and returns focus to the Menu button.

  Validation:

  - npm run lint passes.
  - npm run build-js passes.
  - Existing build warnings remain unchanged.

Testing:
 - I have tested on Chrome only.

### DIFF
--- a/repository-data/webfiles/src/main/resources/site/src/js/header/global-header-escape-close.js
+++ b/repository-data/webfiles/src/main/resources/site/src/js/header/global-header-escape-close.js
@@ -1,0 +1,20 @@
+const menuButtonSelector = '#nhsd-global-header__menu-button';
+
+function isMenuOpen(menuButton) {
+    return menuButton.getAttribute('aria-expanded') === 'true';
+}
+
+export default function initGlobalHeaderEscapeClose() {
+    document.addEventListener('keydown', (event) => {
+        if (event.key !== 'Escape') {
+            return;
+        }
+
+        const menuButton = document.querySelector(menuButtonSelector);
+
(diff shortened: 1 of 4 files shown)


GeoffreyHayward:
Has been superseded by https://github.com/NHS-digital-website/hippo/pull/2861

PR #2870, opened 2026-07-21
[WSP-1089] Check Publication Param before using

GeoffreyHayward:


sankar-shunmuga (approved):
thanks @GeoffreyHayward for the PR. approved.

PR #2869, opened 2026-07-15
[WSP-1088] Fix Feed Hub query params

GeoffreyHayward:

PR #2887, opened 2026-08-17
[WSP-1088] Fix Feed Hub query params

GeoffreyHayward:
Same as approved https://github.com/NHS-digital-website/hippo/pull/2869 but into freeze-trunk-for-post-v16.


pawlozam (approved):
Approved, see Geoff's comment

